PR #4, opened 2016-11-02
Publish settings and others

juanpedromoreno:
Fixes https://github.com/47deg/freestyle/issues/1

Temporarily, Snapshots artifacts are being published at https://oss.sonatype.org/content/repositories/snapshots/com/fortysevendeg/freestyle_2.11/, since we do not have permissions to write at `io.freestyle` domain.

Please @raulraja could you review? Thanks!


raulraja (approved):
LGTM!

jdesiloniz:
@juanpedromoreno as of right now there are no artifacts in sonatype, container folders are published for scala and sjs but not any files inside. Could there be anything wrong with our current publish settings?

PR #8, opened 2016-11-27
freestyle-fetch integration

raulraja:
@dialelo @peterneyens I realize I was going in the wrong direction regarding the approach with persistence and in this case Fetch.
I think this is a compromise that would allow integrations to simply integrate in the monadic/applicative flow of the app exposing a way to go to any `M[_]` as long as the integration supports it. In this case Fetch supports many and thanks to it it becomes a great example of integrating into an app for which multiple target `M[_]` are supported in this case Task, Future, and whatever can conform to `FetchMonadError`.
This integration is based on the ability of a lib exposing ops to lift those to `FreeS` and `FreeS.Par` implementing `FreeSLift`. All instances automatically receive syntax to do it.

In the case of Fetch we expose the run ops as an algebra so Fetch ops are also interpreted when the entire program is interpreted.

Let me know your thoughts and if there is something you'd change.


ghost:
Amazing! Really nice to be able to interleave Fetch with your own algebras 👏 👏

peterneyens:
:+1:

I think there are a couple of unused imports, but they can be cleaned up later.

PR #9, opened 2016-11-29
Add doobie interop

peterneyens:
- Add doobie interop based on the fetch interop.
- Clean up some imports

Only the (in progress) Config tests seem to fail.


raulraja:
This is awesome and I'm sorry I did not realized I was in master when I push some changes late last night. I'll get that fixed!

PR #10, opened 2016-11-29
Logging integration

franciscodr:
This pull request adds logging stuff into freestyle to be integrated in the monadic/applicative flow of the app.

@raulraja @peterneyens @dialelo Any thoughts? Let me know if I should change anything


peterneyens:
I think we somehow need to specify the `C` type, currently it is inferred as `Nothing`:

```
15:13:45.366 [Journal-logger-daemon] DEBUG scala.runtime.Nothing$ - Message
```

I am not sure what the simplest solution is :thinking:

franciscodr:
Yes, I used `val log = Logger[this.type]` in a first approach, but the logger always pointed out to the interpreter.

In this case, type inference seems to not be working well. Maybe @raulraja could shed some light on this problem

raulraja:
Macro :trollface:

raulraja:
A macro is the only way to know what the call site is unless explicitly passed into eh operation arguments AFAIK

franciscodr:
@raulraja I have introduced the changes that you asked me for. Could you take a look? Thanks!

peterneyens:
I don't think we can use journal with ScalaJS ?

ghost:
> I don't think we can use journal with ScalaJS ?

probably not, and also the tests for ScalaJS are not running as far as i can see.

juanpedromoreno:
What about using https://github.com/jokade/slogging? Seems like it would be compatible with Scala/JVM and Scala.js.

franciscodr:
Yes, I took a look at this library this afternoon. I can give it a chance tomorrow. My only concern is that I'm not sure if this library works in an asynchronous way

juanpedromoreno:
Right, seems like journal uses its own thread pool, so it doesn't block the thread where the code is running. https://github.com/jokade/slogging hasn't that feature so you might have two options:

* Use journal for the JVM logging and implements a ConsoleLogger compatible with Scalajs (on your own or even using slogging or any other logging library for Scalajs).
* Use https://github.com/jokade/slogging or equivalevent already compatible with Scalajs.

raulraja:
This is a good example where different interpreter may be used based on the
platform. +1 to @juanpedromoreno suggestion

freestyle-logging/jvm/src/main/scala/io/freestyle/loggingJVM.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object loggingJVM {
+
+  object implicits {
+    implicit def freeStyleLoggingInterpreter[M[_], C: Manifest](implicit ME: MonadError[M, Throwable]): LoggingM.Interpreter[M] =
+      new LoggingM.Interpreter[M] {
+
+        val logger = Logger[C]
+
+        def debugImpl(msg: String): M[Unit] = ME.catchNonFatal(logger.debug(msg))
```
raulraja, line 15: Trying to understand why we need a MonadError instance to wrap all these calls. Also it may be nice to have the rationale in comments added to the interpreter for future reference.

raulraja (changes requested):

freestyle-logging/jvm/src/main/scala/io/freestyle/loggingJVM.scala
franciscodr: @raulraja Maybe I'm wrong but the return type of all the Logger methods is `Unit`. If we want to integrate the logging stuff into a program monadic flow, we should lift the return value into the applicative functor.

Initially I was using `Monad.pure` but I changed to `MonadError.catchNonFatal` after talking to you.

raulraja:
Sounds good, we'll reevaluate if needed in the future once we have a better type for capturing effects.

PR #23, opened 2017-01-03
Added Traversable effect

raulraja:
The traversable effect allows traversables to be lifted to a monadic contex so they can act similar to generators. Constrains are to have a `Foldable` instance and their target runtime to support `MonadCombine`. @franciscodr @peterneyens @juanpedromoreno can you please review? `.freeS` syntax still not provided will come in separate PR to revamp syntax for all effects and other built in algebras.


freestyle-effects/shared/src/main/scala/effects/traverse.scala
```diff
@@ -0,0 +1,44 @@
...
+     * in scope
+     */
+    object implicits {
+      implicit def freestyleTraverseMInterpreter[F[_], M[_]](
+          implicit MC: MonadCombine[M],
+          FT: Traverse[G]): TraverseM.Interpreter[M] =
+        new TraverseM.Interpreter[M] {
+          def emptyImpl[A]: M[A]           = MC.empty[A]
+          def singletonImpl[A](a: A): M[A] = MC.pure[A](a)
+          def fromTraversableImpl[A](ta: G[A]): M[A] = {
```
peterneyens: I think that this should be more efficient than the combination of `traverse` and `unite` :

```scala
val monoidMA = MonoidK[M].algebra[A] 
Foldable[G].foldMap(ta)(Applicative[M].pure)(monoidMA)

// or shorter
val monoidMA = MC.algebra[A] 
ta.foldMap(MC.pure)(monoidMA)
```

Which also means that you only need `Foldable[G]` (instead of `Traverse[G]`), I think ?

raulraja: Sounds great, thanks!

PR #24, opened 2017-01-03
Added `liftFS` syntax for state and traverse effects

raulraja:
@peterneyens Can you please review? thanks!

### DIFF
--- a/freestyle-effects/shared/src/main/scala/effects/traverse.scala
+++ b/freestyle-effects/shared/src/main/scala/effects/traverse.scala
@@ -14,9 +14,9 @@ object traverse {
     /** Acts as a generator providing traversable semantics to programs
      */
     @free sealed abstract class TraverseM[F[_]] {
-      def empty[A]: FreeS[F, A]
-      def singleton[A](a: A): FreeS[F, A]
-      def fromTraversable[A](ta: G[A]): FreeS[F, A]
+      def empty[A]: FreeS.Par[F, A]
+      def singleton[A](a: A): FreeS.Par[F, A]
+      def fromTraversable[A](ta: G[A]): FreeS.Par[F, A]
     }
 
     /** Interpretable as long as Foldable instance for G[_] and a MonadCombine for M[_] exists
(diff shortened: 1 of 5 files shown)


freestyle-effects/shared/src/main/scala/effects/traverse.scala
```diff
@@ -32,6 +32,14 @@ object traverse {
           def fromTraversableImpl[A](ta: G[A]): M[A] = ta.foldMap(MC.pure)(MC.algebra[A])
         }
     }
+
+    class TraverseFreeSLift[F[_]: TraverseM] extends FreeSLift[F, G] {
+      def liftFSPar[A](fa: G[A]): FreeS.Par[F, A] = TraverseM[F].fromTraversable(fa)
+    }
+
+    implicit def freeSLiftStateInspect[F[_]: TraverseM]: FreeSLift[F, G] =
```
peterneyens: This should probably be something like `freeSLiftTraverse` ?

peterneyens:
Apart of that implicit function name, LGTM :+1: 

If you want to add similar syntax for `ReaderM#reader`, I think we might get a conflict with `StateM#inspect` if someone wants to use both `ReaderM` and `StateM` ?

raulraja:
@peterneyens thanks for bringing this up, I realize while testing the potential solution that the module annotation fails to compile when generating Coproducts for modules with dependent types like reader and state. Working on fixing that first and will add it all to this same PR

raulraja:
@peterneyens I added support for nested `@module` containing references to algebras that are defined path dependent and aggregated in the `@module` which was a bug arising from trying the syntax collision. I see your point about the classing syntax though. For now I don't plan to support syntax for `reader` but would love to change the syntax if you can think of an alternative way to do this. 

It seems as if this may be an implicit lookup conflict with no useful error messages when using colliding syntax from more than one algebra in the same for comprehension. Even if I explicitly add the syntax to each one of `reader` and `state` implicits and I bypass the typeclass based definition, scala is still unable to find the syntax when both algebras are intermixed but finds them when separate. Any ideas as to how to improve syntax for all cases is welcome!

PR #25, opened 2017-01-05
Added splain for better implicit error reporting

raulraja:
@fedefernandez Can you please review?


fedefernandez (approved):
I wanna see it in action, LGTM!

PR #49, opened 2017-01-23
jorgegalindo-remove-test-information

jorgegalindocruces:
Hi freestylers, 

I just deleted a few lines that is not in use into this Readme file.

Can you check it @raulraja?


raulraja:
LGTM, thanks for your contribution!

PR #54, opened 2017-01-30
Update cats to 0.9 and others

guersam:
The build fails on my machine due to https://github.com/sbt/sbt/issues/2786.

Possible solutions:
- Apply the [workaround](https://github.com/sbt/sbt/issues/2786#issuecomment-258906054)
- Update to the [latest scalafmt plugin](https://github.com/olafurpg/scalafmt/pull/610) which dropped the reformat on compile setting

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -81,7 +81,7 @@ lazy val freestyle = (crossProject in file("freestyle")).
   settings(name := "freestyle").
   settings(
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-free" % "0.8.1",
+      "org.typelevel" %%% "cats-free" % "0.9.0",
       "org.scala-lang" % "scala-reflect" % scalaVersion.value
     )
   ).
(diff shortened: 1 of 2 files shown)


guersam:
Travis seems comfortable with it anyway :roll_eyes:

raulraja:
@guersam First of all thanks for your contribution!. I got your fork running fine in my machine and was able to `+clean` and `+test` and it works both in 2.11 and 2.12. What is that makes it fail for you but work on Travis and for me locally? I want to make sure before we merge this that others won't see issues building locally in their machines. 

As for a workaround I think upgrading scalafmt and creating a custom sbt task to keep the current behavior would be an acceptable path going forward. Let me know your thoughts and thanks again for taking care of this. Cheers!

guersam:
Hmm, it works on my desktop which runs Ubuntu 16.10 now. The failing environment was my laptop, but it has almost the same setup with the desktop has. I guess it's caused by an outdated global sbt plugin and it won't be that problematic in general.

Using fresh new scalafmt seems great anyway. How about filing another issue for this and keep the current version for now?

raulraja:
@guersam Makes sense, added https://github.com/47deg/freestyle/issues/56. Thanks for your contribution!

PR #68, opened 2017-02-20
Added pcplod test (WIP)

### LINKED ISSUE #67: ensime/scala-ide support: i.e. tests with pcplod
You don't need to do anything special for it to "just work" in ensime and scala-ide, but you might find that you have bugs that break position information when the compiler enables them. I can try to help you fix it if the tests don't pass.

To ensure that you cover this use-case, please make sure to add a pcplod unit test https://github.com/ensime/pcplod

btw, I'm really looking forward to when this is ready for release, so I can use it in
https://github.com/fommil/drone-dynamic-agents/blob/master/src/main/scala/algebra.scala !

raulraja:
Fixes #67 

Proves generated macro code works in the scalac presentation compiler. 
Reported in #67

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,56 +4,6 @@ addCommandAlias("debug", "; clean ; test")
 
 addCommandAlias("validate", "; +clean ; +test; makeMicrosite")
 
-onLoad in Global := (Command.process("project freestyle", _: State)) compose (onLoad in Global).value
-
-val dev  = Seq(Dev("47 Degrees (twitter: @47deg)", "47 Degrees"))
-val gh   = GitHubSettings("com.fortysevendeg", "freestyle", "47 Degrees", apache)
-val vAll = Versions(versions, libraries, scalacPlugins)
-
-lazy val commonSettings = Seq(
-  scalaVersion in ThisBuild := "2.12.0",
-  crossScalaVersions in ThisBuild := Seq("2.11.8", "2.12.0"),
-  scalaOrganization in ThisBuild := "org.typelevel",
-  organization in ThisBuild  := gh.org,
-  organizationName in ThisBuild  := gh.publishOrg,
-  homepage := Option(url("http://www.47deg.com")),
-  organizationHomepage := Some(new URL("http://47deg.com")),
-  startYear := Some(2016),
-  description := "A Cohesive & Pragmatic Framework of FP centric Scala libraries",
-  scalacOptions in ThisBuild ++= Seq(
-    "-Ypartial-unification", // enable fix for SI-2712
-    "-Yliteral-types",       // enable SIP-23 implementation
-    "-Xplugin-require:macroparadise",
-    "-deprecation",
-    "-encoding", "UTF-8",
-    "-feature",
-    "-language:existentials",
-    "-language:higherKinds",
-    "-language:implicitConversions",
-    "-language:reflectiveCalls",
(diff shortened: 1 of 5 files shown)


fommil:
you have a few typos, and your sbt build is a bit messy... I got it to the point where 

```
tests/testOnly * -- -z presentation
```

fails with what looks like a real error

```
SEVERE: RangePosition(pcplodtest.scala, 248, 287, 291): value exec is not a member of pcplodtest.PcplodTestAlgebra[pcplodtest.PcplodTestAlgebra.T]
List(PcMessage(pcplodtest.scala,Error,value exec is not a member of pcplodtest.PcplodTestAlgebra[pcplodtest.PcplodTestAlgebra.T]))
[info] - should generated code works in the presentation compiler *** FAILED ***
[info]   List(PcMessage(pcplodtest.scala,Error,value exec is not a member of pcplodtest.PcplodTestAlgebra[pcplodtest.PcplodTestAlgebra.T])) was not a empty (test.scala:441)
```

and I'm sure once this is fixed up you can see how to add additional assertions

fommil:
see https://github.com/fommil/freestyle/tree/rr-fix-67-test-pcplod

raulraja:
@fommil Got the test to pass without any macro changes. Once I got the syntax right it was just missing the typical implicits imports.If you have a chance to take a look I'd appreciate it. Thanks!

fommil:
sweet! Can you please add a few assertions around your code? e.g. that type at point works in a few places and so on. I assume this code example already shows that the generated code is being generated?

Incidentally, I had imagined defined an algebra using a sealed trait with classes underneath, and the methods would get generated. This is the other way round, I wasn't expecting that! It might be better this way...

raulraja:
@fommil Added more assertions. Travis seem to fail with a RuntimeException in 2.12 because:
```
java.lang.RuntimeException: Conflicting cross-version suffixes in: org.scala-lang.modules:scala-xml
```
Could that be related to the changes in the sbt file?

fommil:
you might want to look into adding a travis rule to only build pushes to master

build.sbt
```diff
@@ -248,19 +188,28 @@ lazy val freestyleLoggingJS  = freestyleLogging.js
...
+    fork in Test := true,
+    javaOptions in Test ++= {
+      val options = (scalacOptions in Test).value.distinct.mkString(",")
+      val cp = (fullClasspath in Test).value.map(_.data).filter(_.exists()).distinct.mkString(",")
+      Seq(
+        s"""-Dpcplod.settings=$options""",
+        s"""-Dpcplod.classpath=$cp"""
+      )
+    },
+    dependencyOverrides += "org.scala-lang" % "scala-compiler" % scalaVersion.value
```
juanpedromoreno: Do we actually need this `dependencyOverrides` setting? I guess this configuration is already set up from the `projectSettings` (AutoPlugin), isn't it?

juanpedromoreno (approved):
I added a minor comment about a potentially duplicated settings, other than that, LGTM.

build.sbt
fommil: yeah, that looks like a dupe.

You need way more `dependencyOverrides` than you'd think, because Ivy is dumb.

raulraja:
@fommil Thanks for your help setting this up and thanks for pcplod, it's awesome!

Issue #37, opened 2017-01-11
Cache Algebras


raulraja:
We should provide a simple algebra with the most common operations for caching. This may be able to reutilize `StateM` generic in the target type of the cache storage so we can build storages for common integrations such as Redis and other related libs in the caching space.

diesalbla:
These are a few extra notes on the ticket: 

We should define an algebra for basic operations of a cache, or any Key-Value store, such as `get`, `put`, `delete`, or others. Then we should create two implementations for this algebra:
* One implementation is backed by a concurrent in-memory map, such as [Java `ConcurrentHashMap`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html). This implementation would work both in the JVM and JavaScript.
* Another implementation is based on a [`Redis`](https://redis.io/) client, with the data being stored in a `Redis` server. The parameters for the connection are to be fetched through the `freestyle-config` module.

PR #70, opened 2017-02-21
FS-37 Cache Algebra And Implementations

diesalbla:
This PR fulfills ticket #37. We introduce two new `sbt` modules, `freestyle-cache` and `freestyle-cache-redis`, that define an algebra of _cache_ operations, and two interpreters for it. In this ticket, we understand by _cache_ an efficient, high-load, and potentially parallel, key-value store. 

#### Algebra 
In the `sbt` project `freestyle-cache`, in the package `freestyle.cache`, we define an algebra named `@free trait CacheM[F[_]]`, that describes a type-class of carrier types `F[_]` that support _cache_ operations. This algebra gives the usual operations for Key-Value store: `get`, `put`, `delete`, `hasKey`, `keys` (for key-set), `clear`. 
Because we want the `CacheM` trait to be generic on the type of indexing keys and on the type of stored   values, we wrap the `@free` algebra inside an empty `KeyValueProvider[K,V]` class. Besides the algebra we provide an `implicit` method to build an interpreter for the algebra (with the same `K` and `V` types). This interpreter is built out of two pieces. 
* One is a _raw_ interpreter to an intermediary carrier type `F[_]`, as an object that inherits from trait  `KeyValueMap[F,K,V]`. The reason for this is that in some implementations the raw format `F` is fixed.
* A natural transformation from the raw format `F` to the desired one `G`.

#### HashMap
In the same project, in the package `freestyle.cache.hashmap`, we provide an implementation for an interpreter of the algebra, in which we use as a store a [`ConcurrentHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) from the `java.util` library. 

* Hasher: the `ConcurrentHashMap` is based on the use of the methods `hashcode` and `equals`, which unfortunately is not always properly implemented for all types. We introduce a type-class `Hashable`, to indicate that the `Key` type must provide a hashing method.

#### Redis
In the project `freestyle-cache-redis`, in the package `freestyle.cache.redis`, we implement an interpreter for the algebra based on an external Redis data-store. For the implementation, we use the Redis client library [`rediscala`](https://github.com/etaty/rediscala/), which is based in `akka`. 

In this interpreter, the notion of _parallelization_ is based on the use of Redis transactions, which for our purposes allows us to group a _sequence_ of data-independent operations and get them sent and executed atomically by the redis server. This parallelization is implemented using the `Applicative` instance of `Kleisli`, where the `Kleisli` is used to defer the communications to the client. 

@raulraja ¿Could you have it reviewed?


freestyle-cache/shared/src/main/scala/cache.scala
```diff
@@ -0,0 +1,117 @@
...
+      def put(key: Key, newVal: Val): FreeS.Par[F, Unit]
+
+      // Removes the entry for the key if one exists
+      def del(key: Key): FreeS.Par[F, Unit]
+
+      // Returns whether there is an entry for key or not.
+      def has(key: Key): FreeS.Par[F, Boolean]
+
+      // Returns the set of keys in the store
+      def keys: FreeS.Par[F, Seq[Key]]
```
raulraja: Since the default `Seq` imported in Predef is mutable and pattern matching is easier over `List` can this return something else that is not `Seq`?

freestyle-cache/shared/src/main/scala/hashmap/Hasher.scala
```diff
@@ -0,0 +1,21 @@
+package freestyle.cache.hashmap
+
+trait Hasher[A] {
+  def hashCode(a: A): Int
+}
+
+object Hasher {
```
raulraja, line 7: We can include here automatic instances for any product type such as case classes, `HList`, `TupleN` etc with shapeless generic derivation given there is also an implicit `Hasher` for the contained elements.

freestyle-cache/shared/src/test/scala/CacheTests.scala
```diff
@@ -0,0 +1,127 @@
+package freestyle.cache
+
+import cats.{~>, Applicative, Id}
+import cats.syntax.cartesian._
+import cats.instances.option._
+import freestyle._
+import freestyle.implicits._
+import org.scalatest._
+
+class CacheTests extends WordSpec with Matchers with CacheTestContext {
```
raulraja, line 10: Nice :clap:

freestyle-cache/shared/src/main/scala/cache.scala
```diff
@@ -0,0 +1,117 @@
...
+     * We assume that the actual store is too big or remote to allow for general
+     * operations over all values, or to search or to iterate over all keys
+     */
+    @free sealed trait CacheM[F[_]] {
+
+      // Gets the value associated to a key, if there is one */
+      def get(key: Key): FreeS.Par[F, Option[Val]]
+
+      // Sets the value of a key to a newValue.
+      def put(key: Key, newVal: Val): FreeS.Par[F, Unit]
```
raulraja, line 21: I would add the following methods if we can support them both on the InMemory and Redis impl.
```scala
def putAll(keyValues: Map[Key, Val]): FreeS.Par[F, Unit]
def putIfAbsent(key: Key, newVal: Val): FreeS.Par[F, Unit]
def replace(key: Key, newVal: Val): FreeS.Par[F, Unit]
def isEmpty: FreeS.Par[F, Boolean]
def containsKey(key: Key): FreeS.Par[F, Boolean]
def containsValue(v: Val): FreeS.Par[F, Boolean]
```

freestyle/shared/src/main/scala/freestyle/Capture.scala
```diff
@@ -5,6 +5,13 @@ import simulacrum.typeclass
 
 import scala.util.Try
 
+/*
```
raulraja, line 8: Thanks!

raulraja (approved):
@diesalbla Excellent work! This is very exciting! Added a few ideas but we can add a new ticket for those improvements :+1: :tada:

freestyle-cache-redis/shared/src/main/scala/rediscala/RedisMap.scala
```diff
@@ -22,13 +22,13 @@ class MapWrapper[M[_], Key, Value](
...
-    RediscalaCont.del(Seq(key)).transform(toM).void
+    RediscalaCont.del(List(key)).transform(toM).void
 
   override def hasKey(key: Key): Ops[M, Boolean] =
     RediscalaCont.exists(key).transform(toM)
 
-  override def keys: Ops[M, Seq[Key]] =
-    RediscalaCont.keys.transform(toM).map(_.map(parser).flatten)
+  override def keys: Ops[M, List[Key]] =
+    RediscalaCont.keys.transform(toM).map(_.map(parser).flatten.toList)
```
raulraja: Can this be simplified to `flatMap`?

freestyle-cache/shared/src/main/scala/cache.scala
diesalbla: The methods `putAll` and `putIfAbsent` are straight-forward. AAMOF, the latter is directly supported in [redis](https://redis.io/commands/setnx). The command `isEmpty` can be supported efficiently through the use of the [`scan` redis command](https://redis.io/commands/scan). The `containsValue` is the one that could be less efficient to implement.

In any case, this can be left to a low-hanging-fruit ticket.

raulraja: Sounds good, since you seem to have more insight, would you mind creating one and tagging it as such? thanks! @anamariamv Is getting started with Freestyle and she could take on that task as well.

PR #71, opened 2017-02-22
Integrate FS2 Streams into FreeS programs

### LINKED ISSUE #61: Integrate FS2 Streams into freestyle's monadic flow
Research the possibility to include [FS2](https://github.com/functional-streams-for-scala/fs2) streams into freestyle's monadic flow. Similarly to the integration with Fetch, we can reify fs2 stream operations into an algebra an defer execution to the free monad interpretation.

ghost:
A first approach to fs2 Stream integration into FreeS programs, using a cats free monad as the stream's effect type.

When running a `StreamM` computation, we go from a  `Stream` to the cats free monad, and then foldmap cats free to the target monad. It supports signaling errors in streams.

I'm also investigating whether I can embed other effect types like `Suspendable`.

Closes #61

### DIFF
--- a/freestyle-fs2/shared/src/main/scala/fs2.scala
+++ b/freestyle-fs2/shared/src/main/scala/fs2.scala
@@ -59,5 +59,13 @@ object fs2 {
           s.runLast.run.foldMap(attemptF)
       }
     }
+
+    implicit class Fs2FreeSyntax[A](private val s: Stream[Eff, A]) extends AnyVal {
+      def liftFS[F[_]](implicit MA: Monoid[A], SF: StreamM[F]): FreeS[F, A] =
+        liftFSPar.freeS
+
+      def liftFSPar[F[_]](implicit MA: Monoid[A], SF: StreamM[F]): FreeS.Par[F, A] =
+        SF.runFold(MA.empty, MA.combine)(s)
+    }
   }
 }
--- a/freestyle-fs2/shared/src/main/scala/fs2.scala
+++ b/freestyle-fs2/shared/src/main/scala/fs2.scala
@@ -1,0 +1,63 @@
+package freestyle
+
+import _root_.fs2._
+import _root_.fs2.util.{Attempt, Catchable, Free}
+
+import cats._
+import cats.free.{Free => CFree}
+import cats.arrow.FunctionK
+
+import scala.concurrent._
+
+object fs2 {
+  type Eff[A] = Free[CFree[Attempt, ?], A]
+
+  implicit val catsFreeC: Catchable[CFree[Attempt, ?]] = new Catchable[CFree[Attempt, ?]] {
+    def pure[A](a: A): CFree[Attempt, A]                              = CFree.pure(a)
+    def attempt[A](fa: CFree[Attempt, A]): CFree[Attempt, Attempt[A]] = fa.map(Attempt(_))
+    def fail[A](err: Throwable): CFree[Attempt, A]                    = CFree.liftF(Left(err))
+    def flatMap[A, B](a: CFree[Attempt, A])(f: A => CFree[Attempt, B]): CFree[Attempt, B] =
+      a.flatMap(f)
+  }
+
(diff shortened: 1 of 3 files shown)


freestyle-fs2/shared/src/main/scala/fs2.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @free sealed trait StreamM[F[_]] {
+    def run[A](s: Stream[Eff, A]): FreeS.Par[F, Unit]
+    def runLog[A](s: Stream[Eff, A]): FreeS.Par[F, Vector[A]]
+    def runFold[A, B](z: B, f: (B, A) => B)(s: Stream[Eff, A]): FreeS.Par[F, B]
+    def runLast[A](s: Stream[Eff, A]): FreeS.Par[F, Option[A]]
+  }
+
+  object implicits {
```
raulraja, line 37: We should include here an instance for `FreeSLift` so we can have syntax to embed streams directly without mixing in the `StreamM` algebra. Here is an example on how that is done in the `Fetch` impl: 
```scala
class FetchFreeSLift[F[_]: FetchM] extends FreeSLift[F, Fetch] {
      def liftFSPar[A](fetch: Fetch[A]): FreeS.Par[F, A] = FetchM[F].runA(fetch)
    }

    implicit def freeSLiftFetch[F[_]: FetchM]: FreeSLift[F, Fetch] = new FetchFreeSLift[F]
```

freestyle-fs2/shared/src/test/scala/Fs2Tests.scala
```diff
@@ -0,0 +1,101 @@
...
+    "allow a stream to be consumed for its last element inside a program monadic flow" in {
+      val program = for {
+        a <- app.nonStream.x
+        b <- app.streamM.runLast(Stream.emits(List(1, 2, 41)))
+      } yield b.fold(0)(_ + a)
+
+      program.exec[Future] map { _ shouldBe 42 }
+    }
+
+    "allow a stream to be run for its effects inside a program monadic flow" in {
```
raulraja, line 59: Once we have the syntax implemented we should also add a test for it to ensure implicits work and the so does the syntax

raulraja (changes requested):
@dialelo this looks awesome!, I added a couple of comment to enable direct syntax for use cases where we can avoid mixing in the `StreamM` algebra

freestyle-fs2/shared/src/main/scala/fs2.scala
ghost: The signature of `liftFSPar` can't be implemented with `Stream` since the only way to turn a `Stream[_[_], A]` into `A` is to fold over the stream.

```scala
    class Fs2FreeSLift[F[_]: StreamM] extends FreeSLift[F, Stream[Eff, ?]] {
      def liftFSPar[A](s: Stream[Eff, A]): FreeS.Par[F, A] = ???
    }
```

raulraja: In this case I can think of an alternative not actually implementing `FreeSLift` since it's too restrictive and given is a fold we can only have syntax when `A` has a Monoid instance.  I think something like this as part of the `implicits` object where the interpreter is now would work to allow `Stream[Eff, A]` to get lifted to `FreeS.Par[F, A]`. Thoughts?
```scala
implicit class FS2FreestyleStreamSyntax[A]
        (private val s: Stream[Eff, A]) extends AnyVal {
      def liftFS[F[_]](implicit MA: Monoid[A], SF: StreamM[F]): FreeS[F, A] =
        liftFSPar[F].freeS
      def liftFSPar[F[_]](implicit MA: Monoid[A], SF: StreamM[F]): FreeS.Par[F, A] =
        SF.runFold[A, A](MA.empty, MA.combine)(s)
    }
```

ghost: Neat! Just added a commit including this.

raulraja (approved):

Issue #72, opened 2017-02-22
Add docs about Cache


anamariamv:
We need to add docs that explain cache, implemented in #37

PR #79, opened 2017-02-27
72 Cache Docs

diesalbla:
This PR closes ticket #72.

We add to the documentation a section about the `Cache` effect. This section contains the following: 

* A description about the `CacheF` effect algebra. We describe its type parameters, and its operations. For these operations, we give an intuition about each of them and a few algebraic properties that specify their semantics.

* A description of its handlers/interpreters, the type classes they require, and the patterns they use.

### DIFF
--- a/docs/src/main/resources/microsite/data/menu.yml
+++ b/docs/src/main/resources/microsite/data/menu.yml
@@ -37,6 +37,9 @@ options:
       - title: Traverse
         url: docs/effects/#Traverse
 
+      - title: Cache
+        url: docs/effects/Cache/
+
   - title: Patterns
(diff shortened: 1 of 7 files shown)


ghost:
Outstanding work @diesalbla, these docs are a joy to read. My only nitpick: maybe we want to add a section explaining how to implement your own `CacheM` interpreter from scratch? For instance taking an existing implementation like ConcurrentHashMap and breaking it down; I think it can help people writing their own interpreters. 

Other than that LGTM

ghost (approved):

raulraja:
Excellent docs @diesalbla!

PR #86, opened 2017-03-07
87 - Code Improvements to the `@free` macro

diesalbla:
This PR carries a few changes to the implementation of the `@free` macro, to make it easier to debug or port.

One change is the use of nomenclature: 
* The parameteric `trait Foo[F[_]]` or `abstract class Foo[F[_]]` provided by the user, which is annotated by the `@free` annotation, is referred to as an _effect_.
* Each method in this _Effect_ whose return type is a `FreeS[F, A]` or `FreeS.Par[F, A]` is called a _Request_. The ADT (`trait` and `case classes`) used to represent these methods is called the _request ADT_. 
* The `class` that builds elements of the `Adt` using the `cats.free.Inject` instance is called the _Lifter_. 
* The `trait` that matches the objects in the Request ADT and issues operation in the domain `F` is called the _Handler_.

We also reorganize the code: we wrap the functionalities using a [macro bundle](http://docs.scala-lang.org/overviews/macros/bundles.html), and we use an internal class `Request` that, from the request `def` provided as input in the `@free` trait or abstract class (the effect), it gives all the elements related to that request.

@raulraja ¿Could you review?

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/free.scala
@@ -42,12 +42,22 @@ object free {
     def mkRequestTrait(requestTypeName: TypeName) =
       q"sealed trait $requestTypeName[A] extends Product with Serializable"
 
-      def mkRequestClass(effectTyParams: List[TypeDef], requestType: TypeName, sc: DefDef): ClassDef = {
-        val retType = sc.tpt.asInstanceOf[AppliedTypeTree].args.last
-        val tparams = effectTyParams.tail ++ sc.tparams
-        val args    = sc.vparamss.flatten //.filter(v => !v.mods.hasFlag(Flag.IMPLICIT))
-        //TODO user trait type params need to be added to the case class leave generated because its args may contain them
-        val className = classNameFor(sc)
+
+      /* A Request declaration in an Effect Trait, such as
+       *
+       * @free trait UserRepository[F[_]] {
+       *     def get(id: Long): FreeS[F, User]
+       *
+       * gets translated to a Request class such as
+       *
+       *     case class Get(id: Long) extends UserRepositoryOp[User]
+       */
+      def mkRequestClass(effectTyParams: List[TypeDef], requestType: TypeName, reqDef: DefDef): ClassDef = {
+        val retType = reqDef.tpt.asInstanceOf[AppliedTypeTree].args.last
+        // Note: Effect trait type params are added to the request case class because its args may contain them
+        val tparams = effectTyParams.tail ++ reqDef.tparams
+        val args    = reqDef.vparamss.flatten //.filter(v => !v.mods.hasFlag(Flag.IMPLICIT))
+        val className = classNameFor(reqDef)
         args match {
           case Nil =>
             q"""case class $className[..$tparams]() extends $requestType[$retType] """
@@ -57,7 +67,7 @@ object free {
       }
 
     def mkSmartCtor(
-      effectTrait: ClassDef,
+      effectName: TypeName,
       effectTyParams: List[TypeDef],
       requestType: TypeName,
       requestDefs: List[DefDef],
@@ -69,7 +79,6 @@ object free {
         requestDefs.zip(requestClasses)
           .map(p => mkSmartCtorImpl(requestType, cpType, p._1, p._2) )
 
-      val effectName  = effectTrait.name.toTermName.toTypeName
       val implName    = TypeName( effectName.decodedName.toString + "_default_impl")
       val firstTParam = effectTyParams.head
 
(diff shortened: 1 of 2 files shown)


freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -11,208 +11,163 @@ class free extends StaticAnnotation {
...
-      }, clsParams, userTrait)
-    }
+    // OP is the name of the Root trait of the Effect ADT
+    lazy val OP = TypeName("T")
+    // MM Is the target of the Interpreter's natural Transformation
+    lazy val MM = TypeName("MM")
+    // LL is the target of the Lifter's Injection
+    lazy val LL = TypeName("LL")
+    // LIFT is the name of the Lifter's Injector Class: no sense in separate name
+    lazy val LIFT = TypeName("Lift")
```
raulraja: I think referring to the smart constructors as `Lifter` or `Lift` is confusing since we already have https://github.com/47deg/freestyle/blob/master/freestyle/shared/src/main/scala/freestyle/Lift.scala.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -11,208 +11,163 @@ class free extends StaticAnnotation {
...
         }
-        companionApply = adtLeaf match {
-          case c: ClassDef => q"${adtLeaf.name.toTermName}[..${c.tparams.map(_.name)}](..$args)"
-          case _           => q"new ${adtLeaf.name.toTypeName}"
+
+      }
+
+      def handlerDef: DefDef =
+        if (params.isEmpty)
+          q"protected[this] def $reqImpl[..$tparams]: $MM[$Res]"
```
raulraja, line 70: For a while I though about making the abstract methods in the synthetic Interpreter not to ending in `Impl` such as `xImpl` and be just `x`. Thoughts? I think it would be a good change to introduce when we potentially get rid of the `[F[_]]` parameterization requirement

raulraja (changes requested):
@diesalbla Looks great, left a few minor comments. I'm concerned about calling the smart constructors `Lifter` I think we should find a better name as we already have `FreeSLift` and `Lift.scala` which is used in every module for entirely different purpose. We should rename that to something different so it does not create confusion down the road for new devs joining the project.

freestyle/shared/src/main/scala/freestyle/free.scala
diesalbla: Given that the _effect_ trait (the one with the `@free` annotation) and the interpreter are part of a different class hierarchy, there should be no problems in their methods having the same names. And since the _Injection_ and the _Interpreter_ are to be defined in different contexts, there is no much room for confusion.

freestyle/shared/src/main/scala/freestyle/free.scala
diesalbla: I was not sure as to what would be the right word for it. I thought of the word `Injector`, but I did not like it much. It could also be _InjectTo_, which makes sense since the first param is the functor it is being injected into, e.g.
```Scala
object Foo {
    class InjectTo[ LL[_] ] {
        def bar(x: Int) : FreeS.Par[LL, Int]
    }
``` 
Other names could be _RaiseTo_, as opposed to calling the `Interpreter` as a  _Handler_, although that has not been renamed.

diesalbla:
From this comments, we are going to rename `Interpreter` as `Handler`, the `Lifter` as `RaiseTo`, and the `xxxImpl` methods of the interpreter as just `xxx`. This has to be changed at several other modules.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -11,208 +11,163 @@ class free extends StaticAnnotation {
...
-      val userTrait @ ClassDef(clsMods, clsName, clsParams, clsTemplate) = cls.duplicate
-      if (!clsMods.hasFlag(Flag.TRAIT | Flag.ABSTRACT))
-        fail(s"@free requires trait or abstract class")
-      mkCompanion(clsName.toTermName, clsTemplate.filter {
-        case _: DefDef => true
-        case _         => false
-      }, clsParams, userTrait)
-    }
+    // OP is the name of the Root trait of the Effect ADT
+    lazy val OP = TypeName("T")
```
raulraja, line 37: I thought we were going to call the root trait just `Op` literally. I'd like to get rid of the type aliases if we can as I'm having issues finishing the IntelliJ macro support impl because of type aliases. Can we do this in this PR so I can test the plugin with all the new changes and ensure what I'm emitting works after this refactor?https://gitter.im/JetBrains/intellij-scala?at=58bfd7331465c46a56014ac5

raulraja (approved):

diesalbla:
After discussion, I think that renaming the `trait Interpreter`, the type `T` for the ADT root, and the methods `xxxImpl` within the `Interpreter`, is a matter for a different issue.

diesalbla:
This PR is now described in relation to issue #87.

Issue #88, opened 2017-03-08
Rename elements in the `@free` macro


diesalbla:
The scope of this ticket is to rename several elements from the code generated by the core `@free` macro annotation, which are redefined 

* The inner type alias `T`, for the root of the ADT of the operations, as `Op`, which is a bit more informative.
* The `trait Interpreter[F[_]]` as a `trait Handler[ F[_]]`, which indicates that this trait's responsability is to _handle_ the effect requests raised by the operations provided by the `@free` trait. 
* The operations in the `Interpreter` (or `Handler`) from `xxxImpl` to `xxx`. Since the `@free` trait's methods and those of the `Interpreter` are not part of a same hierarchy, there is no risk of collisions. 

This issue is opened for discussing these changes, initially proposed in PR #86.

diesalbla:
Regarding the interpreter-handle, as an example, for the following trait: 
```Scala
@free sealed trait Foo[ F[_] ]{
    def ann(x: Int): FreeS.Par[F, String]
}
```
Before the changes, we would have the following 
```Scala
class Bar[F[_] ] extends Foo.Interpreter[F] {
    def annImpl(x: Int): F[String]
}
```
afterwards, it would be 
```Scala
class Bar[F[_] ] extends Foo.Handler[F] {
    def ann(x: Int): F[String]
}
```
This would propagate to all files, since these signatures are used in them.

diesalbla:
@peterneyens @dialelo Opinions welcome.

PR #90, opened 2017-03-08
88 Rename Free Elements

diesalbla:
This PR completes #88. We also update the documentation on the `@free` annotation.


raulraja (approved):
570795b874f4708042016165752

Issue #85, opened 2017-03-06
Multiple provider style effects conflict


peterneyens:
It seems that when you want to use two provider style effects the implicits are not resolved.

For example `effect.reader` and `effect.writer` (with respectively `EnvironmentProvider` and `AccumulatorProvider`) : 

```scala
val rd = reader[Config]
val wr = writer[Vector[String]]

@module trait App[F[_]] {
  val persistence: Persistence[F]

  val errorM: ErrorM[F]
  val readerM: rd.ReaderM[F]
  val writerM: wr.WriterM[F]
}
```

Results in something like : 

```
diverging implicit expansion for type FunctionK[rd.ReaderM.T, Stack]
```

PR #92, opened 2017-03-09
Add an example using a common stack as target

### LINKED ISSUE #32: Provide common stack target example
We need to provide an example of interpreting to a common stack such as `Kleisli[Task, Config, A]` that supports most of the most common effects.

peterneyens:
Currently contains two versions of the example :

1. one using `error`, `reader` and `writer`
2. one using `error`, `reader` and `cache`

The second one is probably the best since the target stack is simpler and no additional `MTL` type class instances are needed, what do you think ?

This additionally fixes the divergent implicit expansion when multiple provider style effects are used or when `Inject` instances are used for nested algebras in the REPL / tut by using `shapeless.Lazy`.

This means that freestyle will also depend on shapeless after merging this PR.

Fixes #32 and #85.


peterneyens:
After choosing which version we include, I also need to add a link in the menu.
Any preference where? Maybe between _effects_ and _patterns_?

raulraja:
Let's add this as a top level item in the menu before patterns

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+import modules._
+import cats.data.{NonEmptyList, ValidatedNel}
+
+def validateOrder[F[_]](order: Order, customer: Customer)(implicit app: App[F]): FreeS.Par[F, ValidatedNel[String, Unit]] =
+  app.readerM.reader { config =>
+    val v = ().validNel[String]
+    v.ensure(NonEmptyList.of(
+      "Number of crates ordered should be bigger than zero"))(
+      _ => order.crates > 0) |+|
```
raulraja, line 96: Since new users may not be familair with `|+|` we should add some text in the description of this paragraph that briefly explain that it comes from cats and what it's purpose is.

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+    v.ensure(NonEmptyList.of(
```
raulraja, line 97: Same for `NonEmptyList` as mentioned for `|+|`

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+We will create a couple of domain specific exception case classes.
+
+```tut:book
+sealed abstract class AppleException(val message: String) extends Exception(message)
+case class CustomerNotFound(id: CustomerId)               extends AppleException(s"Customer $id can not be found")
+case class QuantityNotAvailable(error: String)            extends AppleException(error)
+case class ValidationFailed(errors: NonEmptyList[String]) extends AppleException(errors.intercalate("\n"))
+```
+
+We can use the `validateOrder` method in combination with our persistence algebras and the `error` and `writer` effect algebras, to create the `processOrder` method tying everything together.
```
raulraja: We should explain the core concepts of what the `error` and `writer` algebras are and how they are being used here for fail fast error handling and accumulation

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+```tut:book
+def processOrder[F[_]](order: Order)(implicit app: App[F]): FreeS[F, String] =
+  for {
+    customerOpt <- app.persistence.customer.getCustomer(order.customerId)
```
raulraja: instead of having long lines of `app.persistence.customer.getCustomer` can we have instead those imported so the for comprehension has as little noise as possible?

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+                       QuantityNotAvailable(
+                         s"""There are not sufficient crates of ${order.cultivar} apples in stock
+                            |(only $nbAvailable available, while ${order.crates} needed).""".stripMargin)
+                     )
+                   )
+    _          <- app.persistence.stock.registerOrder(order)
+  } yield s"Order registered for customer ${order.customerId}"
+```
+
+As target type of our computation we choose a combination of `Kleisli` (also known as `ReaderT`), `WriterT` and `fs2.Task`.
```
raulraja: We should briefly explain why we choose this stack and what each one of the main monads stacked here resolve from the point of view of application concerns.

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+```tut:book
+import freestyle.effects.error.implicits._
+import wr.implicits._
+import rd.implicits._
+import KleisliMTL._
+
+val result: Stack[String] = program.exec[Stack]
+```
+
+We can run our `Stack` by supplying a `Config` value, running the `WriterT` and running the `Task`.
```
raulraja: We should end with a conclusion of the advantages of using Freestyle here to create programs descriptions that are interpretable and how evaluation and effects are pushed toward the edge etc. It may not be immediately obvious to someone why this may be a better approach over imperatively writing the same logic or even using vanilla scala.

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+```tut:book
+val cultivars = Set("granny smith", "jonagold", "boskoop")
+val config = Config(cultivars)
+
+val task: Task[(Vector[String], String)] = result.run(config).run
```
raulraja: Add code comments or description of what this does

docs/src/main/tut/docs/stack/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+task.unsafeRunSync.fold(println, println)
```
raulraja: Add code comments or description of what this does

docs/src/main/tut/docs/stack/cache.md
```diff
@@ -0,0 +1,241 @@
+---
+layout: docs
```
raulraja: Comments on the review of the previous file regarding the example, explanations etc.. should also apply on this one.

docs/src/main/tut/docs/stack/cache.md
```diff
@@ -0,0 +1,241 @@
...
+When validating an order we need to get the information of the customer, if a customer places multiple orders we don't want to send a database request every time, so we could use the `freestyle-cache` module.
+
+In the `getCustomer` function we try to find the customer in the cache, otherwise falling back to getting the customer from a database (or other persistence store).
+
+```tut:book
+import cats.data.OptionT
+
+def getCustomer[F[_]](id: CustomerId)(implicit app: App[F]): FreeS[F, Option[Customer]] =
+  // first try to get the customer from the cache
+  OptionT(app.cacheM.get(id).freeS).orElseF {
```
raulraja: Can we replace here `OptionT` with the `OptionM` effect algebra to avoid using a transformer? One of the nicest thigns about these effects algebras is that for most common cases they remove the need to use transformers in the program and logic declarations since the constrains are in the interpreters.

raulraja (changes requested):
Excellent examples @peterneyens! Added a few comments regarding more context in some of the used types and improvements for readability by folks that are not acquainted with cats or FP en Scala to have a lower barrier entry.

docs/src/main/tut/docs/stack/cache.md
peterneyens: The `OptionT` is here only used as a convenient temporary wrapper over `FreeS[Option[T]]` since this method also returns `FreeS[Option[Customer]]`.

The same code without `OptionT` would be : 

```scala
app.cacheM.get(id).flatMap { 
  case s @ Some(_) =>
    (s: Option[Customer]).pure[FreeS[F, ?]]
  case None =>
    app.persistence.customer.getCustomer(id).map {
      case s @ Some(cust) =>
        app.cacheM.put(id, cust)
        s
      case none =>
        none
    }
}
```

I don't think that is clearer.
I also don't think we could use `OptionM` in this case, because we don't actually want to put the `Option` in the stack.

raulraja: got it. Read that way too fast.

raulraja (approved):
@peterneyens this is excellent. Great job!. Not sure if we want to refactor this to include the new `ValidationM` algebra if it fits or you'd rather do that down the road. We didn't have that when we started but that is now it's in master and it may make sense to include it here to further simplify it if it makes sense to you. I'll leave this open and feel free to merge at will.

Issue #83, opened 2017-03-03
ValidationM Effect Algebra


raulraja:
In the same way the `ErrorM` effect implements a fail fast strategy that immediately stops computation upon an error and relies on the presence of a `MonadError` instance for `M[_]` in the interpreter it should possible to have an algebra for Validation which places similar constrains.

In contrast to `ErrorM`, `ValidationM` should be able to accumulate errors and rely in something like `MonadState` in the final interpreter to accumulate errors over the computation and make them available at any point in the program.

The `ValidationM` algebra may tentatively look something like this:

```scala
trait ValidationProvider[E, A] {
  @free trait ValidationM[F[_]] {
       def valid(a: A): FreeS.Par[F, A] //lifts a valid value in the validation context
       def invalid(a: A): FreeS.Par[F, Unit] //accumulates an error in the validation context
       def validation: FreeS.Par[F, ValidatedNel[E, A]] // return the current state of the validation context
       def errors: FreeS.Par[F, List[E]] // returns a list of errors from the validation context
  }
}
```

raulraja:
Other useful combinators in this algebra may be automatically lifting into the validation state the following values:
* from Option[A] 
* from Either[E, A] 
* from ValidatedNel[E, A]

peterneyens:
I am not sure you could combine this `ValidationM` effect with the `StateM` effect if it is also based on `MonadState`, because I don't think you can have multiple `MonadState` constraints for different kind of states.

---

Example :

```scala
import cats.implicits._
import cats.MonadState
import cats.data.{State, StateT}

type ValidationIntState[A] = StateT[State[Int, ?], List[String], A]

MonadState[ValidationIntState, List[String]]
// cats.MonadState[ValidationIntState,List[String]] = cats.data.StateTInstances$$anon$3@56c6d53d

MonadState[ValidationIntState, Int]
// error: could not find implicit value for parameter F: cats.MonadState[ValidationIntState,Int]
//        MonadState[ValidationIntState, Int]
```

PR #94, opened 2017-03-10
Validation effect

ghost:
Addresses #83 

As @peterneyens pointed out in the issue, this algebra requires the free program to be interpreted to a state monad that keeps track of the list of errors. I've tried a few approaches to see if I could relax this requirement but haven't found one.

The combinators allow for accumulating errors by hand, or from `Either` and `ValidatedNel` instances. I didn't include `fromOption` because there will never carry an error that we want to remember. Custom errors can be accumulated by implementing the `ValidationException` trait.

### DIFF
--- a/freestyle-effects/shared/src/main/scala/effects/validation.scala
+++ b/freestyle-effects/shared/src/main/scala/effects/validation.scala
@@ -3,7 +3,7 @@ package freestyle.effects
 import freestyle._
 import cats.{MonadState}
 import cats.data.{State, Validated, ValidatedNel, NonEmptyList}
-import cats.arrow.{FunctionK}
+
 
 object validation {
   /** A validation exception with an explanation. **/
--- a/freestyle-effects/shared/src/main/scala/effects/validation.scala
+++ b/freestyle-effects/shared/src/main/scala/effects/validation.scala
@@ -1,14 +1,17 @@
 package freestyle.effects
 
 import freestyle._
-import cats.{Monad, MonadState, ~>}
-import cats.data.{State, StateT, Validated, ValidatedNel, NonEmptyList}
+import cats.{MonadState}
+import cats.data.{State, Validated, ValidatedNel, NonEmptyList}
 import cats.arrow.{FunctionK}
 
 object validation {
+  /** A validation exception with an explanation. **/
   trait ValidationException {
     def explanation: String
   }
+
+
   case class NotValid(explanation: String) extends ValidationException
 
   object ValidationException {
@@ -17,12 +20,16 @@ object validation {
 
   type Errors = List[ValidationException]
 
+  /** An algebra for introducing validation semantics in a program. **/
   @free sealed trait ValidationM[F[_]] {
     def valid[A](x: A): FreeS.Par[F, A]
+
     def invalid(err: ValidationException): FreeS.Par[F, Unit]
+
     def errors: FreeS.Par[F, Errors]
 
     def fromEither[A](x: Either[ValidationException, A]): FreeS.Par[F, Either[ValidationException, A]]
+
     def fromValidatedNel[A](x: ValidatedNel[ValidationException, A]): FreeS.Par[F, ValidatedNel[ValidationException, A]]
   }
 
--- a/freestyle-effects/shared/src/main/scala/effects/validation.scala
+++ b/freestyle-effects/shared/src/main/scala/effects/validation.scala
@@ -6,58 +6,50 @@ import cats.data.{State, Validated, ValidatedNel, NonEmptyList}
 
 
 object validation {
-  /** A validation exception with an explanation. **/
-  trait ValidationException {
-    def explanation: String
-  }
+  final class ValidationProvider[E]{
+    type Errors = List[E]
 
+    /** An algebra for introducing validation semantics in a program. **/
+    @free sealed trait ValidationM[F[_]] {
+      def valid[A](x: A): FreeS.Par[F, A]
 
-  case class NotValid(explanation: String) extends ValidationException
+      def invalid(err: E): FreeS.Par[F, Unit]
 
-  object ValidationException {
-    def apply(msg: String): ValidationException = NotValid(msg)
-  }
+      def errors: FreeS.Par[F, Errors]
 
-  type Errors = List[ValidationException]
+      def fromEither[A](x: Either[E, A]): FreeS.Par[F, Either[E, A]]
 
-  /** An algebra for introducing validation semantics in a program. **/
-  @free sealed trait ValidationM[F[_]] {
-    def valid[A](x: A): FreeS.Par[F, A]
+      def fromValidatedNel[A](x: ValidatedNel[E, A]): FreeS.Par[F, ValidatedNel[E, A]]
+    }
 
-    def invalid(err: ValidationException): FreeS.Par[F, Unit]
+    object implicits {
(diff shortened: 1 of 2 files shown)


freestyle-effects/shared/src/main/scala/effects/validation.scala
```diff
@@ -0,0 +1,63 @@
...
+
+    def fromEither[A](x: Either[ValidationException, A]): FreeS.Par[F, Either[ValidationException, A]]
+
+    def fromValidatedNel[A](x: ValidatedNel[ValidationException, A]): FreeS.Par[F, ValidatedNel[ValidationException, A]]
+  }
+
+  object implicits {
+    implicit def freeStyleValidationMStateInterpreter[M[_]](
+      implicit
+        MS: MonadState[M, Errors]
```
raulraja: @peterneyens Would this face the same issues described in https://github.com/47deg/freestyle/issues/85 ?

freestyle-effects/shared/src/main/scala/effects/validation.scala
```diff
@@ -0,0 +1,63 @@
+package freestyle.effects
+
+import freestyle._
+import cats.{MonadState}
+import cats.data.{State, Validated, ValidatedNel, NonEmptyList}
+
+
+object validation {
+  /** A validation exception with an explanation. **/
+  trait ValidationException {
```
raulraja: We may consider here wrapping the entire Algebra in a provider similar to how the `ReaderM` and `StateM` effects work so that `ValidationException` can be any `E` provided by the user. The issue with us providing a unsealed datatype which people need to extend from is that once they unwrap the errors and wish to coalesce errors into something else the compiler won't be able to do exhaustive checking that all cases are contemplated.
For example at evaluation time:
```scala
program[ValidationM.Op].exec[Logger].run(List.empty) map { case (errors, a) => 
   errors match {
     case Nil => ...
     case h :: _ => h match {
        case CustomExceptionExtendingValidationException =>
        case _ =>  // Requiered for exahustive matching because ValidationException is open to extension
     }
   }
}
```

freestyle-effects/shared/src/main/scala/effects/validation.scala
```diff
@@ -0,0 +1,63 @@
...
+
+      def fromEither[A](x: Either[ValidationException, A]): M[Either[ValidationException, A]] =
+      x match {
+        case Left(err) => MS.flatMap(invalid(err))((unit) => MS.pure(x))
+        case Right(_) => MS.pure(x)
+      }
+
+      def fromValidatedNel[A](x: ValidatedNel[ValidationException, A]): M[ValidatedNel[ValidationException, A]] =
+        x match {
+          case Validated.Invalid(errs: NonEmptyList[ValidationException]) => MS.pure(x)
```
peterneyens: The `MS.pure(x)` here is probably a copy paste error, no ?

freestyle-effects/shared/src/test/scala/effects/EffectsTests.scala
```diff
@@ -223,6 +222,102 @@ class EffectsTests extends AsyncWordSpec with Matchers {
...
+    }
+
+    "fromValidatedNel" in {
+      import cats.data.{Validated, ValidatedNel, NonEmptyList}
+
+      def program[F[_]: ValidationM] =
+        for {
+          a <- ValidationM[F].fromValidatedNel(Validated.Valid(42))
+          b <- ValidationM[F].fromValidatedNel(
+            Validated.Invalid(NonEmptyList(MissingFirstName, List.empty)) : ValidatedNel[ValidationException, Unit]
```
peterneyens: I think we can write this as `Validated.invalidNel[ValidatedException, Unit](MissingFirstName)`.

freestyle-effects/shared/src/test/scala/effects/EffectsTests.scala
```diff
@@ -223,6 +222,102 @@ class EffectsTests extends AsyncWordSpec with Matchers {
...
+
+    "valid" in {
+      def program[F[_]: ValidationM] =
+        for {
+          _ <- Applicative[FreeS[F, ?]].pure(1)
+          b <- ValidationM[F].valid(42)
+          _ <- Applicative[FreeS[F, ?]].pure(1)
+        } yield b
+
+      program[ValidationM.Op].exec[Logger].run(List.empty) map {  _ shouldBe Tuple2(List(), 42) }
```
peterneyens: FWIW, we could replace these `.run(List.empty)` with `.runEmpty` if we `import cats.instances.list._`.

raulraja (approved):
@dialelo Excellent work! I made some comments regarding parameterization of error values which I find interesting to support proper Errors modeled as ADTs.
We may also consider here method extensions for any type  `A` in the implicits section
```scala
a.invalid : FreeS.Par[F, Unit] // lifts a as an error
a.valid : FreeS.Par[F, A] // lifts as a a valid value
```

freestyle-effects/shared/src/main/scala/effects/validation.scala
peterneyens: I think my `shapeless.Lazy` changes in #92 should solve #85.

freestyle-effects/shared/src/main/scala/effects/validation.scala
ghost: 👍 Did it

freestyle-effects/shared/src/main/scala/effects/validation.scala
ghost: Good catch!

freestyle-effects/shared/src/test/scala/effects/EffectsTests.scala
ghost: Done!

freestyle-effects/shared/src/test/scala/effects/EffectsTests.scala
ghost: Much more clear, done 👍

ghost:
Addressed your comments and added also syntax methods as requested by @raulraja, thanks for the review!

raulraja (approved):
Awesome and great work!

freestyle-effects/shared/src/main/scala/effects/validation.scala
```diff
@@ -0,0 +1,62 @@
...
+              ))((unit) => MS.pure(x))
+            case Validated.Valid(_) =>  MS.pure(x)
+          }
+      }
+
+      implicit class ValidSyntax[A](private val s: A){
+        def valid[F[_] : ValidationM]: FreeS[F, A] = ValidationM[F].valid(s)
+      }
+      implicit class InvalidSyntax[A](private val e: E){
+        def invalid[F[_] : ValidationM]: FreeS[F, Unit] = ValidationM[F].invalid(e)
```
peterneyens, line 56: I am wondering if the `valid` and `invalid` syntax might not be confusing as there is similar syntax in `cats.syntax.validated`?

raulraja: @peterneyens @dialelo we can rename it to whatever else makes sense. `correct`, `wrong` ? or anything along those lines.

ghost: Maybe we could avoid the clash by naming them something like `liftValid` and `liftInvalid`?

PR #98, opened 2017-03-13
Add validation docs

### LINKED ISSUE #97: Add docs for the Validation effect
Now that we've merged the PR that includes validation effet, we should write the docs for it and add it to the "Effects" section.

ghost:
Closes #97 

I've also changed the name of the implicit methods for lifting valid and invalid values to FreeS programs since @peterneyens mentioned that they may lead to confusion with `cats.syntax.validated`.

### DIFF
--- a/docs/src/main/tut/docs/effects/README.md
+++ b/docs/src/main/tut/docs/effects/README.md
@@ -401,3 +401,110 @@ def programEmpty[F[_]: TraverseM] =
 
 programEmpty[TraverseM.Op].exec[List]
 ```
+
+## Validation
+
+The validation effect allows for the distinction between valid and invalid values in a program, accumulating the validation errors when executing it.
+
+The `validation` effect, like `state`, supports parametrization to any type remaining type safe throughout the program declaration.
+
+The constrains placed by this effect is that there needs to be an implicit evidence of `MonadState[M[_], List[E]]`
+for any runtime `M[_]` where `E` is the type of the validation error.
+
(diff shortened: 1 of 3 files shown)


raulraja (approved):
Looks great!

docs/src/main/tut/docs/effects/README.md
```diff
@@ -401,3 +401,125 @@ def programEmpty[F[_]: TraverseM] =
 
 programEmpty[TraverseM.Op].exec[List]
 ```
+
+## Validation
+
+The validation effect allows for the distinction between valid and invalid values in a program, accumulating the validation errors when executing it.
+
+The `validation` effect, like `state`, supports parametrization to any type remaining type safe throughout the program declaration.
```
raulraja: `state` here can be `[state](#state)` so when sbt-microsites https://github.com/47deg/sbt-microsites/issues/74 is fixed local anchor links will take you to the state definition it's referring to.

peterneyens:
The documentation looks OK :+1: 

---

In general about `validation`:

1. I don't know how often monadic error accumulation comes up in practice.

2. I think we should be able to implement `validation` using `MonadWriter` instead of `MonadState` ?
  The `ValidationM` handler could then just interpret into the `WriterM` algebra.
  `WriterM` doesn't seem to have a `listen` operation that makes the accumulated data accessible for the moment though, something we would need for `ValidationM.errors`.

Issue #76, opened 2017-02-22
Extend Cache Operations


diesalbla:
This Issue continues Issue#37, which is under Code Review in PR #70. The goal is to extend the algebra of `CacheM`, with the operations that appear in the original [comment](https://github.com/47deg/freestyle/pull/70#discussion_r102312551).

For the `Interpreter` backed on a `ConcurrentHashMap`, the implementation should be based on the [API](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html) for that class. For the `Redis`-backed interpreter, the implementation should be based on the commands of `redis`, and the client library [`rediscala`](https://github.com/etaty/rediscala/issues). Specially relevant  would be the command [`scan`](https://redis.io/commands/scan) for efficiently iterating on the set of keys.

PR #101, opened 2017-03-13
Extend cache operations

anamariamv:
This PR solves #76 and extend the algebra of Cache, with operations:
```scala
def putAll(keyValues: Map[Key, Val]): FreeS.Par[F, Unit]
def putIfAbsent(key: Key, newVal: Val): FreeS.Par[F, Unit]
def replace(key: Key, newVal: Val): FreeS.Par[F, Unit]
def isEmpty: FreeS.Par[F, Boolean]
````
@raulraja  Could you review, please?

On the other hand, I've added more test in Redis. @diesalbla , Could you review this part, please?

Thanks

### DIFF
--- a/freestyle-cache-redis/shared/src/main/scala/rediscala/Kleislis.scala
+++ b/freestyle-cache-redis/shared/src/main/scala/rediscala/Kleislis.scala
@@ -1,13 +1,10 @@
 package freestyle.cache.redis.rediscala
 
 import cats.data.Kleisli
+
 import scala.concurrent.Future
-import _root_.redis.{ByteStringSerializer => Serializer, ByteStringDeserializer => Deserializer}
-import _root_.redis.commands.{
-  Keys => KeyCommands,
-  Server => ServerCommands,
-  Strings => StringCommands
-}
+import _root_.redis.{Cursor, ByteStringDeserializer => Deserializer, ByteStringSerializer => Serializer}
+import _root_.redis.commands.{Keys => KeyCommands, Server => ServerCommands, Strings => StringCommands}
 
 private[rediscala] trait StringCommandsCont {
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: freestyle-cache/shared/src/main/scala/hashmap/HashMapWrapper.scala [hunk 1]
```diff
@@ -41,6 +42,16 @@ final class ConcurrentHashMapWrapper[F[_], Key, Value](
     ()
   }
 
+  override def putAll(keyValues: Map[Key, Value]): F[Unit] = C.capture {
+    keyValues.map { case (k, v) => table.put(hkey(k), v) }
```
peterneyens: If we replace `map` here with `foreach`, we don't need the `()` on the next line.

freestyle-cache/shared/src/main/scala/cache.scala
```diff
@@ -20,6 +20,12 @@ package cache {
       // Sets the value of a key to a newValue.
       def put(key: Key, newVal: Val): FreeS.Par[F, Unit]
 
+      // Copy all of the mappings from the specified map to this map
```
peterneyens: Should we say `to this cache` instead of `to this map`?

Similarly for `isEmpty` below (`this cache contains` vs `this map contains`)?

raulraja:
Awesome stuff! 
giphy

raulraja (approved):

freestyle-cache-redis/shared/src/main/scala/rediscala/Kleislis.scala
```diff
@@ -26,6 +23,17 @@ private[rediscala] trait StringCommandsCont {
   def set[Key: Format, Value: Serializer](key: Key, value: Value): Ops[Future, Boolean] =
     Kleisli((client: StringCommands) => client.set(key, value))
 
+  def mset[Key : Format, Value: Serializer](keyValues: Map[Key, Value])(implicit format: Format[Key]): Ops[Future, Boolean] = {
```
diesalbla: If you add the implicit parameters `format: Format[Key]`, then you don't need to put the constrain `Key: Format` in the declaration of type parameters.

diesalbla (approved):
Minor comment. Good work.

PR #104, opened 2017-03-14
Clear Macro Module

diesalbla:
This PR solves #104. It applies some maintenance and cleanup to the `@module` macro annotation. 

* We remove unused or duplicated code. 
* We simplify naming of internal elements.
* We try to use more restrictive types for inner methods. 
* We unfold and simplify algorithms to recollect type leaves, and to assemble Coproducts. Most relevantly, we reduce the latter from a `scanLeft` to a `map`.

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/module.scala
+++ b/freestyle/shared/src/main/scala/freestyle/module.scala
@@ -1,13 +1,9 @@
 package freestyle
 
-import cats.instances.tuple._
-import cats.syntax.functor._
-
 import scala.annotation.{compileTimeOnly, StaticAnnotation}
 import scala.language.experimental.macros
 import scala.reflect.macros.whitebox
 import scala.reflect.runtime.universe._
-import scala.reflect.runtime.{currentMirror => cm}
 
 trait Modular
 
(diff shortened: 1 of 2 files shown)


freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -28,94 +24,65 @@ object materialize {
...
-                case x =>
-                  fail(
-                    s"found unexpected case building Coproduct: $x with types ${x.map(_.getClass)}")
-              }
-              (pos + 1, newTree)
-          }
 
+    def mkCoproduct(algebras: List[Type]): List[TypeDef] =
+      algebras.map(x => TermName(x.toString) ) match {
+        case Nil => List( q"type Op[A] = Nothing")
```
raulraja, line 57: Can we evaluate a program with this? perhaps we should at this point fail compilation and indicate that a `@module` needs at least 1 algebra contained within.

raulraja (approved):
Great stuff @diesalbla, thanks for cleaning this up!

freestyle/shared/src/main/scala/freestyle/module.scala
diesalbla: As far as possible (I may have made mistakes), I have tried to stick to an equivalent code transformation, so the new code would, but for some changes in the generated code, do the same that previous operations did, and in any case not break anything. 

In this case, the code prior to the PR was generating such an `Op` if there were no algebras. This can be changed, of course.

PR #105, opened 2017-03-14
Add syntax tests for freestyle-effects

peterneyens:
Additionally clean up package so we don't need `import freestyle._` everywhere.


raulraja (approved):
:clap:

peterneyens:
Seems that the Travis branch build now fails for 2.11.8 because of the microsite Kazari changes ?

Issue #58, opened 2017-02-08
IntelliJ IDEA plugin


raulraja:
IDEA does not support macros. IDEA devs expect lib authors to create plugins such as the one for `simulacrum` https://blog.jetbrains.com/scala/2015/10/14/intellij-api-to-build-scala-macros-support/

raulraja:
This is already implemented in https://github.com/47deg/intellij-scala/commit/d774cb34530db9ee235dd5c53dd16575db4689a8 but is currently bloqued by what we find in #93

raulraja:
Also it seems like the current IDEA plugin API has bugs that affect what we are trying to do with Freestyle https://gitter.im/JetBrains/intellij-scala?at=58c68137dd08b4b859c37cb9

raulraja:
We are migrating all macros to scala meta and now we are running into the issue that IDEA scalameta support is broken in cross projects that target scala.js and scala.jvm. I submitted a PR to fix this and tested locally with freestyle scalameta's branch. PR available here https://github.com/JetBrains/intellij-scala/pull/380

BusyByte:
@raulraja was just wondering the status of this?   Was working in IntelliJ and can't even see the Client generated by the macro at all so this would be extremely.

diesalbla:
@BusyByte After this issue https://github.com/frees-io/freestyle/issues/539, we intend to migrate some of the macro annotations, back from scalameta macros (which are deprecated) to the old Scala macro paradise, which is to be an integral part of Scala 2.13. We will then re-examine the support in IntelliJ for the freestyle macros.

For your immediate problems, I am afraid we won't be fixing support in IntelliJ for current `scalameta`-based macro annotations.

PR #116, opened 2017-03-16
FS Further Changes to @free and @module macros

diesalbla:
This commit carries a few further improvements to the implementations
of the @free and @module macros, that should not change their semantics.

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/free.scala
@@ -4,6 +4,10 @@ import scala.annotation.{compileTimeOnly, StaticAnnotation}
 import scala.language.experimental.macros
 import scala.reflect.macros.blackbox
 
+trait FreeModuleLike {
+  type Op[A]
+}
+
 @compileTimeOnly("enable macro paradise to expand @free macro annotations")
 class free extends StaticAnnotation {
   def macroTransform(annottees: Any*): Any = macro free.impl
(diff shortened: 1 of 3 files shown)


freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -4,6 +4,10 @@ import scala.annotation.{compileTimeOnly, StaticAnnotation}
 import scala.language.experimental.macros
 import scala.reflect.macros.blackbox
 
+trait FreeModuleLike {
```
raulraja, line 7: This is also necessary for the IntelliJ integration and instead of calling it `FreeModuleLike` I think we should call it something like `Algebra`.  Unless the IDEA people get the `Inject` api they provide for libs that use macros fixed as described in #58 users will end up extending this trait to get proper syntax highlighting in IDEA like so:
```scala
@free trait MyOps[F[_]] extends Algebra {
   ....
}

MyOps.Op //visible by IntelliJ syntax highlighter when directly referenced and in `Handler` implementations.
```

raulraja (approved):
Minor comment on the top level trait but this looks great, thanks!

freestyle/shared/src/main/scala/freestyle/free.scala
diesalbla: ¿Do you mean that the `@free` macro should also rewrite the `@free`-annotated trait, so as to add `Algebra` to its subclasses? ¿Or do you just mean to rename the trait?

raulraja: Even if the macro adds the extension to the class IntelliJ is still unable to see it so that is not that important at this point since IDEA can not see anything that the macro generates. Not just that but IDEA can't use any definitions in the macro to infer types in their precompilation parser which they use for code completion.

If the user extends it manually though the type alias is also seen as if it were part of the companion when it really is just declared in the trait. But don't worry about this now, if you agree that Algebra is a better name let's change it or we can down the road when the IDEA plugin makes some progress. 

Based on the scala IDEA gitter channel they are working a new more complete API that maybe makes the user extending anything not necessary.

diesalbla: For now I would just leave it as an exclusive concern of the `@macro` implementation. I prefer not to put more _meaningful_ words, for now.

Issue #93, opened 2017-03-09
Attempt to remove `F[_]` parametrization in Algebras and `FreeS` definitions


raulraja:
Discussing with @diesalbla the other day he mentioned that the `F[_]` constrain could be moved away from the trait and type alises definition by changing the way one received implicit instances of the syntatic classes generated by @free and @module. This issue here is to track an attempt of doing that. I'm interested in this because it would affect how #58 is implmented and we need to send a PR to intellij scala with the actual syntax the macro generates. https://github.com/47deg/intellij-scala/commit/d774cb34530db9ee235dd5c53dd16575db4689a8

PR #117, opened 2017-03-16
DO NOT MERGE emove `F` param: proof of concept

diesalbla:
This PR shows a proof of concept for #93. The goal of this ticket is to see if we can remove the `F[_]` parameter from the definition of the `@free` algebras. This PR carries the implementation in the core package, and as an example it shows the result of the change for the `doobie` algebra.

* We introduce in the core `freestyle` package a `Oper` package, to describe operations in a `@free` algebra. This is needed only because the `@free` trait may also contain other elements apart from the operations of that algebra.

* This parameter, which represents the _carrier_ type to which the operations are _injected to_, is now only available through the type `To`. For every `@free` and `@module` trait `Foo` the respective companion object provides a class `Foo.To[F[_]]`, which contains the operations (for `@free`) or the algebras (for `@module`).  

Note: For convenience, this PR is based in another branch, which carries other changes to `@free` and `@macro` that do not alter the semantics of these macros.

### DIFF
--- a/freestyle-cache-redis/shared/src/test/scala/RedisTests.scala
+++ b/freestyle-cache-redis/shared/src/test/scala/RedisTests.scala
@@ -25,7 +25,7 @@ class RedisTests extends AsyncWordSpec with Matchers with RedisTestContext {
   "Redis integration into Freestyle" should {
 
     "allow to interleave one CacheM within the programs monadic flow" in {
-      def program[F[_]: CacheM]: FreeS[F, Int] =
+      def program[F[_]: CacheM.To]: FreeS[F, Int] =
         for {
           a <- Applicative[FreeS[F, ?]].pure(Some(1))
           b <- CacheM[F].put("Joe", 13) *> CacheM[F].get("Joe")
@@ -38,21 +38,21 @@ class RedisTests extends AsyncWordSpec with Matchers with RedisTestContext {
   "GET" should {
 
     "result in None if the datastore is empty" in {
-      def program[F[_] : CacheM]: FreeS[F, Option[Int]] =
+      def program[F[_] : CacheM.To]: FreeS[F, Option[Int]] =
         CacheM[F].get("a")
 
       program[CacheM.Op].exec[Future] map { _ shouldBe None }
     }
 
     "result in None if a different key is set" in {
-      def program[F[_] : CacheM]: FreeS[F, Option[Int]] =
+      def program[F[_]: CacheM.To]: FreeS[F, Option[Int]] =
         CacheM[F].put("a", 0) *> CacheM[F].get("b")
 
       program[CacheM.Op].exec[Future] map { _ shouldBe None }
     }
 
     "result in Some(v) if v is the only set value" in {
-      def program[F[_] : CacheM]: FreeS[F, Option[Int]] =
+      def program[F[_] : CacheM.To]: FreeS[F, Option[Int]] =
         CacheM[F].put("a", 0) *> CacheM[F].get("a")
 
       program[CacheM.Op].exec[Future] map { _ shouldBe Some(0)}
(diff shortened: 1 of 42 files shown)


diesalbla:
For the `doobie` algebra and tests, this is what the macros generate: 

```Scala
  object DoobieM extends scala.AnyRef {
    import cats.arrow.FunctionK;
    import cats.free.Inject;
    import freestyle.FreeS;

    sealed abstract trait Op[A] extends Product with Serializable;
    case class TransactOP[A] extends Op[A] with scala.Product with scala.Serializable {
      <caseaccessor> <paramaccessor> val f: ConnectionIO[A] = _;
    };

    class To[LL[_]] extends scala.AnyRef {
      implicit <paramaccessor> private[this] val I: Inject[Op, LL] = _;
      def transact[A](f: ConnectionIO[A]): FreeS.Par[LL, A] = FreeS.inject[Op, LL](TransactOP[A](f))
    };

    implicit def to[LL[_]](implicit I: Inject[Op, LL]): To[LL] = new To[LL]();
    def apply[LL[_]](implicit ev: To[LL]): To[LL] = ev;
    abstract trait Handler[MM[_]] extends FunctionK[Op, MM] {
      protected[this] def transact[A](f: ConnectionIO[A]): MM[A];
      override def apply[A](fa: Op[A]): MM[A] = fa match {
        case (l @ TransactOP(_)) => transact(l.f)
      }
    }
  };

  object App extends FreeModuleLike {
    val X = coproductcollect.apply(this);
    type Op[A] = X.Op[A];
    class To[F[_]] extends scala.AnyRef {
      implicit <paramaccessor> val nonDoobie: NonDoobie.To[F] = _;
      implicit <paramaccessor> val doobieM: DoobieM.To[F] = _;
    };
    implicit def to[F[_]](implicit nonDoobie: NonDoobie.To[F], doobieM: DoobieM.To[F]): To[F] = new To[F]();
    def apply[F[_]](implicit ev: To[F]): To[F] = ev
  };
}

{
  final class $anon extends scala.AnyRef {
    type Op[A] = Coproduct[DoobieM.Op, NonDoobie.Op, A]
  };
  new $anon()
}
```

raulraja:
@diesalbla this would be a major and excellent improvement toward reducing boilerplate. Kudos!
Regarding the type aliases. We want those to remain compatible with what they represent meaning they still should point to what `FreeS` and `FreeS.Par` are aliasing today if we proceed with this change. Alternatively we can call them just `FS` or `FS.Par` to further reduce the typing necessary but still denote they refer to `FreeStyle`. Again great stuff and this would be awesome if we can get it to work across the board. :clap:

diesalbla:
@raulraja I will soon send the changes for the whole project. One issue I have noticed, though, is that although you remove boilerplate at some places, at other places you may need to add a bit more boilerplate. However, since the boilerplate that you remove is the confusing unnecessary parameter `F[_]` of the algebra, that would be a benefit.

raulraja:
@diesalbla what are the places that would require more boilerplate?

freestyle-cache-redis/shared/src/test/scala/RedisTests.scala
```diff
@@ -181,14 +181,14 @@ class RedisTests extends AsyncWordSpec with Matchers with RedisTestContext {
...
-      def program[F[_] : CacheM]: FreeS[F, Option[Int]] =
+      def program[F[_] : CacheM.To]: FreeS[F, Option[Int]] =
         CacheM[F].put("a", 0) *> CacheM[F].del("b") *> CacheM[F].get("a")
 
       program[CacheM.Op].exec[Future] map { _  shouldBe Some(0) }
     }
 
     "ignore a succeeding PUT on different key" in {
-      def program[F[_] : CacheM]: FreeS[F, Option[Int]] =
+      def program[F[_] : CacheM.To]: FreeS[F, Option[Int]] =
```
diesalbla, line 191: @raulraja Regarding boilerplate: 

Whenever we want to indicate that, to a type parameter `F : * -> *`, we can put operations from the effect algebra, the type-class is no longer the name of the algebra (no parameters), but the `To` class from its companion object.

diesalbla:
I have now extended the changes to more sections of the library.

diesalbla:
I'm closing this PR to redirect it to `master`.

Issue #106, opened 2017-03-14
Default implemention for operation


peterneyens:
`WriterM.tell` is now implemented using `WriterM.writer` (like it is done in `MonadWriter`). This means that no `Tell` operation is created in the `WriterM` algebra and that the [`tell` method](https://github.com/47deg/freestyle/blob/master/freestyle-effects/shared/src/main/scala/effects/writer.scala#L20) of the handler is never used. As a consequence if a `MonadWriter` instance overrides the default implementation, it wouldn't be used.

I don't know if we can somehow specify that a function with an implementation does need an algebra operation and that it is implementation can be used in the default handler?

I am not sure if it is really needed in this case, but it might be handy in some cases for user specific algebras?

Probably something for macro guys @raulraja and @diesalbla.

raulraja:
Can we fix the `WriterM` impl in the mean time to not implement `tell` in terms of `writer` and ensure the interpreter uses the `MonadWriter` instance?

PR #118, opened 2017-03-16
Fix WriterM.tell

peterneyens:
This adds a TellOp in the algebra and the `MonadWriter.tell` function will be used in the handler/interpreter.

Like mentioned in #106.


raulraja (approved):
:+1:

PR #119, opened 2017-03-16
WIP - FS-93 Remove `F[_]` param from `@free` and `@module` annotations. 

diesalbla:
This PR shows a proof of concept for #93. The goal of this ticket is to see if we can remove the `F[_]` parameter from the definition of the `@free` algebras. This PR carries the implementation in the core package, and as an example it shows the result of the change for the `doobie` algebra.

* We introduce in the core `freestyle` package a `Oper` package, to describe operations in a `@free` algebra. This is needed because 1) the `@free`-annotated trait may contain other elements, apart from the effect operations; and 2) there is a desire to differentiate parallel from sequential operations.

* This parameter, which represents the _carrier_ type to which the operations are _injected to_, is now only available through the type `To`. For every `@free` and `@module` trait `Foo` the respective companion object provides a class `Foo.To[F[_]]`, which contains the operations (for `@free`) or the algebras (for `@module`).

Note: For convenience, this PR is based in another branch, which carries other changes to `@free` and `@macro` that do not alter the semantics of these macros.

This PR succeeds #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,17 +13,17 @@ Applications built with Freestyle can be interpreted to any runtime semantics su
 ```scala
 import freestyle._
 
-@free trait Database[F[_]] {
-  def get(id: UserId): FreeS[F, User]
+@free trait Database {
+  def get(id: UserId): OpSeq[User]
 }
 
-@free trait Cache[F[_]] {
-  def get(id: UserId): FreeS[F, User]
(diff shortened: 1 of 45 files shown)


diesalbla:
The work in this PR has been continued in other branches.

Issue #120, opened 2017-03-16
Split tests Module


diesalbla:
The tests module started when there was only two modules, but now that there are many more it makes sense to host them where they belong.

PR #121, opened 2017-03-16
FS-120 Split Tests Module

diesalbla:
This PR fulfills #120.

We move all of the tests from the `tests` module to the `freestyle` module, since they mostly concerned the definitions at the core package. It turns out that most of the tests in the `tests` module only depend on freestyle, and a few test on `monix`, so we also have to add this dependency.


raulraja (approved):

build.sbt
```diff
@@ -40,6 +40,9 @@ lazy val freestyle = (crossProject in file("freestyle")).
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-free" % "0.9.0",
       "org.scala-lang" % "scala-reflect" % scalaVersion.value,
+      "io.monix" %%% "monix-eval" % "2.2.1" % "test",
+      "io.monix" %%% "monix-cats" % "2.2.1" % "test",
```
peterneyens, line 44: Should we move the monix test to `freestyle-monix`?
That way we wouldn't need monix as a test dependency in the core  `freestyle` module?

diesalbla: Test dependencies are not much of a problem, since they are intended not to be packaged in a release. They only concern those who are developing the library. 

For what is worth, a few other things may be scrapped from the core. For instance, the `Monad[Future]` instance, though useful, does not belong there IMHO.

Another day. :man_shrugging:

raulraja: We have that instance there because `Future` comes with the standard lib and while I agree with you that it doesn ot belong there, it's easier for users to keep a single dependency to run programs that can be potentially parallelizable out of the box. Since `Future` is the most abused `IO` like type in scala we expect most Freestyle usage happening around it.

PR #122, opened 2017-03-17
http4s integration

### LINKED ISSUE #113: Http adapter - Http4s Integration
In order to gracefully integrate with `http4s` we should provide appropriate implicit instances that convert a `FreeS` program to a `fs2.Task` both implicit and explicit conversions

In an ideal scenario a user would simply `import freestyle.http.http4s.implicits._` and automatically embed a Freestyle program like so:

```scala
val bids: FreeS[F, Bids] = ???
val bidsService = HttpService {
  case GET -> Root / "bids" =>
    Ok(bids)
}
```
Since we already have a `fs2` module we can provide syntax there `frees.toFS2Task` and then an implicit conversion in the `http-http4s` module that is brought into scope with this module implicits for http4s users to not worry about calling syntax directly.

This feature will allow users to define purely functional programs whose resulting values are transparently returned in Http Responses. Once this feature is implemented appropriate documentation should go under `Integrations -> Http -> http4s`

peterneyens:
Closes #113.

I sort of based the structure on @dialelo's play branch, but we should check so all http integrations follow the same structure.

### DIFF
--- a/freestyle-http-http4s/src/main/scala/http/http4s.scala
+++ b/freestyle-http-http4s/src/main/scala/http/http4s.scala
@@ -1,0 +1,15 @@
+package freestyle
+package http
+
+import cats.{~>, Monad}
+import org.http4s.EntityEncoder
+
+import freestyle.implicits._
+
+object http4s {
+
+  implicit def freeSEntityEncoder[F[_], G[_], A](
+      implicit NT: F ~> G, G: Monad[G], EE: EntityEncoder[G[A]]): EntityEncoder[FreeS[F, A]] =
+    EE.contramap((f: FreeS[F, A]) => f.exec[G])
+
+}
--- a/freestyle-http-http4s/src/test/scala/http/Http4sTest.scala
+++ b/freestyle-http-http4s/src/test/scala/http/Http4sTest.scala
@@ -1,0 +1,90 @@
+package freestyle.http
+
+import org.scalatest.{AsyncWordSpec, Matchers}
+
+import fs2.Task
+import fs2.interop.cats._
+import org.http4s._
+import org.http4s.dsl._
+
+import freestyle._
+import freestyle.implicits._
+import freestyle.http.http4s._
+
+class Http4sTests extends AsyncWordSpec with Matchers {
+  import algebras._
+  import handlers._
+
+  "Http4s Freestyle integration" should {
+
+    def getUser[F[_]: App](id: Long): FreeS[F, User] =
+      App[F].userRepo.get(id)
+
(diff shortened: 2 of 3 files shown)


freestyle-http-http4s/src/test/scala/http/Http4sTest.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  val users: Map[Long, User] = Map(
+    1L -> User("foo"),
+    2L -> User("bar")
+  )
+
+  implicit val userRepoHandler: UserRepository.Handler[Task] = 
+    new UserRepository.Handler[Task] {
+      def get(id: Long): Task[User] =
+        Task.now(users.get(id).getOrElse(User("default")))
```
raulraja, line 86: Ideally this module does not use `fs2.Task` directly but `Capture` instead depending on `freestyle-fs2`. Also I realized `freestyle-fs2` does not include a `Capture` instance. The goal is that these programs run unchanged in all different Http Servers if their handlers are based of `Capture`

freestyle-http-http4s/src/main/scala/http/http4s.scala
```diff
@@ -0,0 +1,15 @@
...
+import cats.{~>, Monad}
+import org.http4s.EntityEncoder
+
+import freestyle.implicits._
+
+object http4s {
+
+  implicit def freeSEntityEncoder[F[_], G[_], A](
+      implicit NT: F ~> G, G: Monad[G], EE: EntityEncoder[G[A]]): EntityEncoder[FreeS[F, A]] =
+    EE.contramap((f: FreeS[F, A]) => f.exec[G])
```
raulraja, line 13: giphy 1

raulraja (approved):
This is awesome!. We should rewire the tests to include a `Capture` instance and fix the fs2 module to also include that instance in it's implicits and add some docs to explain how you can embed programs in http4s seamlessly.

freestyle-http-http4s/src/test/scala/http/Http4sTest.scala
peterneyens: Yeah, we need to look if we can simplify the `freestyle-fs2`, `freestyle-async-fs2` situation (the same for `freestyle-monix` and `freestyle-async-monix`), but I am not sure how since the first only depend on `freestyle` and the second also depend on `freestyle-async`.

raulraja: Y think it's fine if the inner modules depend on each other if transitively they depend in the same version of fs2 and monix and they don't impose additional dependencies that are unnecessary. Ultimately we want the easiest user experience getting it to work.

PR #123, opened 2017-03-17
Twitter Util support

### LINKED ISSUE #114: Http adapter : Twitter Future + Finagle integration
In order to gracefully integrate with `Finagle` we should provide appropriate implicit instances of Twitter's `Future` data type so freestyle program can be interpreted to those.

This integration will yield two modules ` freestyle-twitter-future` and `freestyle-http-finagle`.  
` freestyle-twitter-future` should include appropriate instances to interpret Freestyle programs to Twitter's future including syntax while `freestyle-http-finagle` can provide implicit conversions to run Freestyle programs at the edge of endpoints return values automatically interpreting to Twitter Future.

In an ideal scenario a user would simply `import freestyle.http.finagle.implicits._` and automatically embed a Freestyle program along these lines:
```scala
val bids: FreeS[F, Response[Bids]] = ???
val httpService = new Service[http.Request, http.Response] {
  def apply(req: http.Request): Future[http.Response] =
    bids
}
```

This feature will allow user to define purely functional programs whose resulting values are transparently returned in Http Responses. Once this feature is implemented appropriate documentation should go under `Integrations -> Http -> Finagle` and `Integrations -> Executors -> Twitter Future`

raulraja:
Fixes #114 

Changed scope of this PR to support Twitter Util types as supported by catbird and providing the extra Capture instances until `ApplicativeEval / MonadDefer` make it to cats and then is supported in catbird


freestyle-twitter-future/src/main/scala/future.scala
```diff
@@ -0,0 +1,37 @@
+package freestyle
+package twitter
+
+import com.twitter.util.Future
+import cats.MonadError
+
+class FreestyleTwitterMonad extends MonadError[Future, Throwable] with Capture[Future] {
+
```
peterneyens: `catbird` (a project with cats type class instances for Twitter data types) does override `ap` as you did before (see [catsbird instances `Future`](https://github.com/travisbrown/catbird/blob/master/util/src/main/scala/io/catbird/util/future.scala)).

peterneyens (approved):

peterneyens:
If (or when) the `ApplicativeEval` / `MonadDefer` Cats PR somehow comes through and these instances are added to `catsbird`, we may need to recheck if this `freestyle-twitter-future` module brings something extra.
We could then maybe add a section to the documentation about how you can use Twitter `Future` with freestyle with the examples from the current tests.

peterneyens:
I should take a look later to test the laws for all the `MonadError` instances we provide.

freestyle-twitter-future/src/main/scala/future.scala
raulraja: I removed it because 'ap' was never called. I'm out of a computer today but if you think it makes sense to add it feel free to commit on this branch and we'll get it merged

raulraja:
@peterneyens makes sense. I did not know about catsbird. If those PRs make it to cats we can get rid of this module and describe how to achieve the same functionality in the docs through catsbird

raulraja:
@peterneyens I updated this PR to support all the datatypes supported by catbird in twitter util and added the additional Capture instances. Can you take another look when you have a chance?

peterneyens:
I am not sure we should provide a `Capture[Var]` instance. `Var` looks similar to `MVar` from scalaz and Monix  and I don't think that is something you would use as a target for a computation.

It also seems that it is not possible to create a stack safe `tailRecM` for `Var` which could be a problem with `Free`.

For the rest :+1:

PR #124, opened 2017-03-20
Finch integration

peterneyens:
Provide implicit conversions to `io.finch.internal.Mapper` which is used by `io.finch.Endpoint#apply`.

Because we need conversions for `FreeS` and `FreeS.Par` if we have either an `Id` or a Twitter `Future` handler given a value, a function or a function for an `Endpoint` with an `HList`, there are 12 (`2 x 2 x 3`) conversion methods.


raulraja (approved):
Great work!

raulraja:
Looks like there are some conflicts, we can merge this as soon as they are resolved

Issue #111, opened 2017-03-16
Http adapter: Play Framework Integration


raulraja:
In order to gracefully integrate with `Play Framework` we should provide appropriate implicit instances of their Action composers to transparently use Freestyle programs as return values in Play Framework Controllers without the need to manually executing them to Future. https://www.playframework.com/documentation/2.5.x/ScalaActionsComposition

In an ideal scenario a user would simply `import freestyle.http.playframework.implicits._` and automatically embed a Freestyle along these lines:
```scala
    val bids: FreeS[F, Bids] = ???
    def route(args : String) = FreeSAction { 
        ...
    }
```

This feature will allow user to define purely functional programs whose resulting values are transparently returned in Http Responses. Once this feature is implemented appropriate documentation should go under `Integrations -> Http -> PlayFramework`.

peterneyens:
Added in #125.

PR #125, opened 2017-03-20
Play integration

ghost:
- [x] Don't try to run freestyle-play tests with Scala 2.12

This is an initial approach to #111 

It allows you to use a `FreeS` program that returns an HTTP result as an action. It can handle both programs that depend on the request or not, so we should be able to use this in Scala Exercises.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,15 @@
 language: scala
-scala:
-- 2.11.8
-- 2.12.1
-jdk:
-- oraclejdk8
+
 before_install:
 - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then bash
   scripts/decrypt-keys.sh; fi
 - export PATH=${PATH}:./vendor/bundle
+
 install:
 - rvm use 2.2.3 --install --fuzzy
 - gem update --system
 - gem install sass
 - gem install jekyll -v 3.2.1
-script:
-- sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
-- sbt ++$TRAVIS_SCALA_VERSION "docs/tut"
 
 after_success:
 - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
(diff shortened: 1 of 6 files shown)


raulraja (approved):
Awesome! Once the tests issue with the scala version is fixed we can merge this.

ghost:
I haven't had success trying to exclude the `freestylePlay` sbt project from running with Scala 2.12. I've tried [this plugin](https://github.com/sbt/sbt-doge) and fiddling with `sbt` settings but I finally settled for changing the Travis configuration to run the full suite for 2.11, and only the `tests` integration project under 2.12.

raulraja:
@juanpedromoreno can you check the new Travis configuration and if it works merge this? Thanks!

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
+  include:
+   - scala: 2.11.8
+     jdk: oraclejdk8
+     script:
+      - sbt ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport
+      - sbt ++$TRAVIS_SCALA_VERSION "docs/tut"
+   - scala: 2.12.1
+     jdk: oraclejdk8
+     script:
+      - sbt ++$TRAVIS_SCALA_VERSION clean tests/test
```
juanpedromoreno: Travis file looks good, however:

* For scala 2.12, we are only running this test: https://github.com/47deg/freestyle/blob/master/tests/src/test/scala/test.scala . Maybe changing `sbt ++$TRAVIS_SCALA_VERSION clean tests/test` by `sbt ++$TRAVIS_SCALA_VERSION clean test` would be better? Or maybe creating a test suite excluding the play tests...
* Can we run `tut` also for scala 2.12?

ghost: > Maybe changing sbt ++$TRAVIS_SCALA_VERSION clean tests/test by sbt ++$TRAVIS_SCALA_VERSION clean test would be better? Or maybe creating a test suite excluding the play tests...

The underlying problem is that `sbt` tries to load the Scala 2.11 `freestyle-play` project using Scala 2.12, so I don't know if creating a test suite excluding those would work. If we run `sbt ++$TRAVIS_SCALA_VERSION clean test ` with Scala 2.12 sbt will fail since Play is not available for Scala 2.12.

> Can we run tut also for scala 2.12?

Done

juanpedromoreno:
@dialelo @raulraja The last commit should work, since it's going to run all the tests except the play integration tests when running on scala 2.12.

ghost:
💚 thanks @juanpedromoreno

PR #127, opened 2017-03-21
sbt-org-policies Plugin Milestone 1 Integration

juanpedromoreno:
This PR is a join-effort from @fedefernandez  and I . It integrates the first `sbt-org-policies` milestone, version `0.2.1`.

As a default setting that the plugin is providing, moves the artifacts from `com.fortysevendeg` to `com.47deg`.

Please, @raulraja could you review? Thanks.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,146 +1,101 @@
-import catext.Dependencies._
-import microsites.util.BuildHelper._
-
 addCommandAlias("debug", "; clean ; test")
 
 addCommandAlias("validate", "; +clean ; +test; makeMicrosite")
 
-lazy val micrositeSettings = Seq(
-  micrositeName := "Freestyle",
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -1,146 +1,116 @@
...
-lazy val freestyleEffectsJS  = freestyleEffects.js
+lazy val freestyleEffectsJS = freestyleEffects.js
 
 lazy val freestyleAsync = (crossProject in file("freestyle-async")).
   dependsOn(freestyle).
   settings(name := "freestyle-async").
   settings(
     libraryDependencies ++= Seq(
-      "org.scalatest" %% "scalatest" % "3.0.1"      % "test"
+      %%("scalatest") % "test"
```
raulraja: Should we move this to a common setting so it's only declared once?

raulraja (approved):
This is great, thanks guys, merge at will when ready

PR #129, opened 2017-03-22
Enable sequencing of FreeS

colin-lamed:


freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -0,0 +1,35 @@
...
+    "make monad implicitly available" in {
+      type G[A] = FreeS[SCtors1.Op, A]
+      implicitly[cats.Monad[G]].isInstanceOf[cats.Monad[G]]
+    }
+
+    "enable sequenceU" in {
+      val s = SCtors1[SCtors1.Op]
+      val program = for {
+        a <- List(s.x(1), s.x(2)).sequenceU
+      } yield a
```
peterneyens: If we drop the for comprehension, we can just write :

```scala
val program = List(s.x(1), s.x(2)).sequenceU
```

Or using `traverseU` :

```scala
val program = List(1, 2).traverseU(s.x)
```

freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -0,0 +1,35 @@
...
+class implicitsTests extends WordSpec with Matchers {
+
+  "Implicits" should {
+
+    import algebras._
+    import freestyle.implicits._
+
+    "make monad implicitly available" in {
+      type G[A] = FreeS[SCtors1.Op, A]
+      implicitly[cats.Monad[G]].isInstanceOf[cats.Monad[G]]
```
peterneyens: We could use scalatest's `compile` check here (like I used to check the availablity of implicit http4s `EntityEncoder`s in the [freestyle-http-http4s module](https://github.com/47deg/freestyle/blob/master/freestyle-http-http4s/src/test/scala/http/Http4sTest.scala#L26-L29))

```scala
import cats.Monad
"Monad[FreeS[SCtors1.Op, ?]]" should compile 
```

freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    "enable sequenceU" in {
+      val s = SCtors1[SCtors1.Op]
+      val program = for {
+        a <- List(s.x(1), s.x(2)).sequenceU
+      } yield a
+      program.isInstanceOf[FreeS[SCtors1.Op, List[Int]]] shouldBe true
```
peterneyens: Should we run `program` instead of using `isInstanceOf` ?

Something like 

```scala
import interps._
program.exec[Option] shouldBe (Some(List1, 2))
```

freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    "enable sequence" in {
+      val s = SCtors1[SCtors1.Op]
+      val program = for {
+        a <- List(s.x(1), s.x(2)).sequence[λ[a => FreeS[SCtors1.Op, a]], Int]
```
peterneyens: We could replace `λ[a => FreeS[SCtors1.Op, a]` with `FreeS[SCtors1.Op, ?]`.

Unrelated to this PR, it seems we don't use `-Ypartial-unification` [in the tests](https://github.com/47deg/freestyle/blob/master/build.sbt#L205)? Otherwise I think we could just write `.sequence` here (without explicitly specifying the type with a type lambda).

freestyle/shared/src/test/scala/freestyle/implicits.scala
colin-lamed: The scalatest 'compile' check doesn't seem to play well with kind-projector compiler plugin. I have left the type alias.

colin-lamed:
I have discovered that explicitly defining a Monad/Applicative for FreeS is not required for clients compiling with the `-Ypartial-unification` option.

So PR may be unrequired.

However, clients compiling with Scala 2.11 cannot use this flag.

freestyle/shared/src/main/scala/freestyle/Implicits.scala
```diff
@@ -24,4 +24,9 @@ trait Interpreters {
     Inject.catsFreeRightInjectInstance(I.value)
 }
 
-object implicits extends Interpreters
+trait Monads {
```
raulraja: Can we rename this to `FreeSInstances`? down the road we may want to explicitly add others here.

freestyle/shared/src/main/scala/freestyle/Implicits.scala
```diff
@@ -24,4 +24,9 @@ trait Interpreters {
...
+  implicit def catsFreeMonadForFreeS[F[_]]: Monad[FreeS[F, ?]] =
```
raulraja: Can we rename this instance to `freestyleMonadForFreeS`?

raulraja (approved):
Thanks @colin-passiv! If Partial unification is not available in 2.11 then this instance makes sense for folk depending on Freestyle in Scala 2.11

PR #149, opened 2017-03-28
Add doc section about doobie integration

### LINKED ISSUE #137: Docs: Doobie section
We need to document the Doobie integration and how you can easily interleave queries in FreeS programs with a few examples.

peterneyens:
Fixes #137.

### DIFF
--- a/docs/src/main/tut/docs/integrations/doobie.md
+++ b/docs/src/main/tut/docs/integrations/doobie.md
@@ -1,0 +1,110 @@
+layout: docs
+title: Doobie
+permalink: /docs/integrations/doobie/
+---
+
+# Doobie integration
+
+It is easy to embed a doobie program in a freestyle program using the _freestyle-doobie_ module. This module provides a very simple algebra with only one operation: `transact`. Handlers for this algebra are available if there is a doobie `Transactor` for the target type.
+
+
+The standard freestyle and cats imports:
+
+```tut:silent
+import freestyle._
+import freestyle.implicits._
+
+import cats.implicits._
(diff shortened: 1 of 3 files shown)


docs/src/main/tut/docs/integrations/doobie.md
```diff
@@ -0,0 +1,110 @@
...
+    def subtract(a: Int, b: Int): Task[Int] = Task.now(a - b)
+  }
+```
+
+A program using the `Example` module:
+
+```tut:book
+def example[F[_]: DoobieM](implicit example: Example[F]): FreeS[F, (Person, Int)] =
+  for {
+    person <- getPerson.liftFS[F]  // analogous to example.doobieM.transact(getPerson)
```
peterneyens, line 99: To use the lifting syntax here, I needed to add the `DoobieM` context bound even though we have an implicit `Example`.

Should we look if we can provide implicit conversions from modules to the underlying algebras, so the syntax can be used directly?

raulraja: That would be awesome!

raulraja (approved):

raulraja:
Looks great @peterneyens, I think we need to fix the link in the menu though to point to these docs.

peterneyens:
I have no idea how the test coverage can go down with this PR ?

raulraja:
@peterneyens https://docs.codecov.io/docs/unexpected-coverage-changes They explain why this may be happening but definitely nothing in this PR changing coverage

Issue #131, opened 2017-03-27
Clean up and increase test coverage in preparation for 0.1.0


raulraja:
As we quickly approach 0.1.0 we should increase test coverage to ensure we are above 80% as shown in latest builds by the sbt org policies plugin.

PR #150, opened 2017-03-28
Add tests for all logging operations

peterneyens:
I also dropped the `io.frees` directories from `jvm` and `js` like I did in `shared` before.


peterneyens:
This ~should push~ pushes us over the 80% test coverage threshold mentioned in #131.

raulraja (approved):

PR #151, opened 2017-03-28
Clean up play tests

peterneyens:


raulraja (approved):

peterneyens:
I am not sure why [these two lines](https://github.com/47deg/freestyle/blob/master/http/play/src/main/scala/play.scala#L46-L47) aren't covered.

raulraja:
@peterneyens maybe the test not using that constructor?

PR #152, opened 2017-03-28
Replace isInstanceOf with scalatest `a[]`

peterneyens:
Replaced `isInstanceOf[Foo] shouldBe true` with `shouldBe a[Foo]` like I did in #151 for the play tests and like @fommil mentioned on the gitter channel.


raulraja (approved):
:+1:

PR #153, opened 2017-03-28
Slick Integration

raulraja:
WIP, Do not merge.

Integration ready for review, missing docs will be added once reviewed. Also this depends on `org-policies` 0.3.2-SNAPSHOT waiting for a release from @juanpedromoreno to change to `0.3.2`.


juanpedromoreno:
@raulraja 0.3.2 is already available.

raulraja:
Still getting permgen in 2.11 builds in Travis @juanpedromoreno @fedefernandez but this is ready for review @peterneyens

freestyle-slick/src/main/scala/slick.scala
```diff
@@ -0,0 +1,49 @@
...
+    implicit def freeStyleSlickHandler[M[_]](
+        implicit asyncContext: AsyncContext[M],
+        db: JdbcBackend#DatabaseDef): SlickM.Handler[M] =
+      new SlickM.Handler[M] {
+        def run[A](fa: DBIO[A]): M[A] = asyncContext.runAsync { cb =>
+          db.run(fa).onComplete { t =>
+            cb(t.toEither)
+          }
+        }
+      }
```
peterneyens: Should we provide an explicit `SlickM.Handler[Future]` that just calls `db.run(fa)` ?
That way we wouldn't go from `Future` to async callback and back to `Future` if the target is `Future`.

peterneyens (approved):
I am not 100% familiar with Slick, but this looks good :+1:

freestyle-slick/src/main/scala/slick.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  object implicits {
+    implicit def freeStyleSlickHandler[M[_]](
+        implicit asyncContext: AsyncContext[M],
+        db: JdbcBackend#DatabaseDef): SlickM.Handler[M] =
+      new SlickM.Handler[M] {
+        def run[A](fa: DBIO[A]): M[A] = asyncContext.runAsync { cb =>
+          db.run(fa).onComplete { t =>
+            cb(t.toEither)
```
peterneyens: It seems `Try#toEither` was added when `Either` became right biased in 2.12.
So it seems we need to pattern match on the `Try` : 

```scala
import scala.util.{Success, Failure}
db.run(fa).onComplete {
  case Success(x) => cb(Right(x))
  case Failure(e) => cb(Left(e))
}
```

docs/src/main/tut/docs/slick/README.md
```diff
@@ -0,0 +1,113 @@
...
+    def subtract(a: Int, b: Int): Future[Int] = Future.successful(a - b)
+  }
+```
+
+A program using the `Example` module:
+
+```tut:book
+def example[F[_]: SlickM](implicit example: Example[F]): FreeS[F, (Person, Int)] =
+  for {
+    person <- getPerson.liftFS[F]  // analogous to example.doobieM.transact(getPerson)
```
peterneyens: `example.doobieM.transact` -> `example.slickM.run`.

raulraja:
Still failing with permgen on 2.11 :(

PR #156, opened 2017-03-30
Use fully qualified types for `Product` and `Serializable`

### LINKED ISSUE #155: use fully qualified names when generating code
e.g. you're referring to `Product` and `Serializable` which are defined in `scala._`.

This is important if using the new compiler flag `-Ypredef _` which will not import `scala._` by default.

peterneyens:
Fixes #155.


raulraja (approved):

fommil:
ha! That was easy. Sorry I should really have sent a PR. I've not gotten used to the fact that I can do that sort of thing during working hours now :smile: (previous customer was behind a huge corporate firewall)

PR #157, opened 2017-03-30
Doc section about Cats

### LINKED ISSUE #136: Docs: Cats section
We need to document how you can use Cats imports and machinery to inter-operate with Freestyle declared types. We may also want to expand here with new combinators for the `Frees` companion to sequece, traverse, etc... to simplify usage as we did with `FreeS#pure`

peterneyens:
Resolves #136.


raulraja (approved):
@peterneyens, this is pure gold. thanks so much!

PR #170, opened 2017-04-03
Play integrations reworked as implicit conversion and docs

### LINKED ISSUE #162: Docs: Play Framework integration

raulraja:
Fixes #162

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -253,7 +253,6 @@ lazy val docs = (project in file("docs"))
       %%("fs2-io"),
       %%("fs2-cats"),
       %%("doobie-h2-cats"),
-      %%("play-test"),
       "com.h2database" % "h2" % "1.4.194" % "test"
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -241,6 +241,7 @@ lazy val docs = (project in file("docs"))
   .dependsOn(freestyleCacheJVM)
   .dependsOn(freestyleDoobie)
   .dependsOn(freestyleSlick)
+  .dependsOn(freestylePlay)
   .settings(micrositeSettings: _*)
   .settings(noPublishSettings: _*)
   .settings(
@@ -252,6 +253,7 @@ lazy val docs = (project in file("docs"))
       %%("fs2-io"),
       %%("fs2-cats"),
       %%("doobie-h2-cats"),
+      %%("play-test"),
       "com.h2database" % "h2" % "1.4.194" % "test"
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -200,17 +200,6 @@ lazy val freestyleFs2 = (crossProject in file("freestyle-fs2"))
 lazy val freestyleFs2JVM = freestyleFs2.jvm
 lazy val freestyleFs2JS  = freestyleFs2.js
 
-lazy val freestylePlay = (project in file("http/play"))
-  .dependsOn(freestyleJVM)
-  .settings(name := "freestyle-play")
-  .settings(
-    parallelExecution in Test := false,
-    libraryDependencies ++= Seq(
-      %%("play"),
-      %%("play-test") % "test"
-    )
-  )
-
 lazy val tests = (project in file("tests"))
   .dependsOn(freestyleJVM)
   .settings(noPublishSettings: _*)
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -252,6 +253,7 @@ lazy val docs = (project in file("docs"))
       %%("fs2-io"),
       %%("fs2-cats"),
       %%("doobie-h2-cats"),
+      %%("play-test"),
```
peterneyens: I could be mistaken, but I don't think we actually use _play-test_ in the docs as they are?

docs/src/main/tut/docs/play/README.md
```diff
@@ -0,0 +1,75 @@
...
+  for {
+    msg <- Noop[F].ok
+  } yield Results.Ok(msg)
+```
+
+Once our program is defined we may simply return it as a result of any `Action` without the need to explicitly interpret it to `Future`
+
+```tut:silent
+object AppController extends Controller {
+	def endpoint = Action.async { _ => program[Noop.Op] }
```
peterneyens: Really nitpicking here :smiley:, but we can drop some leading spaces here.

http/play/src/main/scala/play.scala
```diff
@@ -26,26 +26,15 @@ import _root_.play.api.mvc._
 import _root_.play.api.http._
 
 object play {
-  object FreeSAction {
-    def apply[F[_]](prog: FreeS[F, Result])(
-        implicit MF: Monad[Future],
```
peterneyens, line 31: Just wondering: could we loose this implicit `Monad[Future]` parameter if we `import cats.instances.future._`?

docs/src/main/tut/docs/play/README.md
```diff
@@ -0,0 +1,75 @@
+---
+layout: docs
+title: Play Framework
+permalink: /docs/play/
+---
+
+# Play Framework integration
+
+It is easy to a freestyle program as a result of a Play Framework Action with the _freestyle-play_ module. This module provides an implicit conversion from `FreeS[F, A] => Future[A]` which allows a user to define a Free program as the result of any Play Action that expects a Future as a response.
```
peterneyens: This should probably be "It is easy to _use_ a freestyle ..."?

We should probably use either

-  an implicit conversion from `FreeS[F, A]` to `Future[A]` or
-  an implicit conversion `FreeS[F, A] => Future[A]`.

Should we replace the "Free" in "to define a Free program" with `FreeS` or freestyle?

peterneyens (commented):
I left a few comments, but this looks good :+1:

http/play/src/main/scala/play.scala
peterneyens: I apparently commented on the old code, but the same comment holds for the analogous parameter in `freestylePlayFutureConversion`.

raulraja: Good catch!

raulraja:
@peterneyens Addressed all your comments, ready in case you want to take another look.

http/play/src/test/scala/PlayTests.scala
```diff
@@ -60,26 +50,17 @@ class PlayTests extends AsyncWordSpec with Matchers {
...
-      FreeSAction { program[Noop.Op] } shouldBe an[Action[Result]]
+    "FreeS programs can be used as return value in Play actions" in {
+      Action.async { _ =>
+        program[Noop.Op]
+      } shouldBe an[Action[Result]]
     }
 
-    "FreeSAction creates an action from a function that returns a program given a request" in {
-      FreeSAction { request =>
+    "FreeS programs can iinteract with a given request and used as returned values in Play actions" in {
```
peterneyens: Typo `iinteract` -> `interact`

docs/src/main/tut/docs/play/README.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# Play Framework integration
+
+It is easy to use a freestyle program as a result of a Play Framework Action with the _freestyle-play_ module. This module provides an implicit conversion `FreeS[F, A] => Future[A]` which allows a user to define a Free program as the result of any Play Action that expects a Future as a response.
+
+In order to enable this integration you may depend on _freestyle-play_
+
+```scala
+libraryDependencies += "com.47deg" %% "freestyle-play" % "0.1.0"
```
peterneyens: It seems the other http integrations are called `freestyle-http-akka`, `freestyle-http-finch` and `freestyle-http-http4s`.

Should we rename `freestyle-play` to `freestyle-http-play` to be consistent? (or the other way around?)

raulraja: yes, I'll rename it before we merge this

PR #173, opened 2017-04-03
Add Finch and http4s docs

### LINKED ISSUE #160: Docs: Finch integration

peterneyens:
Resolves #160 and #163.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -128,9 +128,7 @@ lazy val freestyleTwitterUtil = (project in file("freestyle-twitter-util")).
   dependsOn(freestyleJVM).
   settings(name := "freestyle-twitter-util").
   settings(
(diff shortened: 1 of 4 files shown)


diesalbla (commented):
The programs and examples seem fine. However, I  really miss a brief explanation about the following: 
* What is the basic piece in each library? What does the programmer have to provide to make a piece. What is its  type? Intuitively, what does it do? ¿Is it a value, or a  function, or a `service`?
* When using `freestyle`, ¿what does the `FreeS` expression compute? ¿What conversion is the `freestyle` integration performing? What does it need to do that?

Should not be too long, given that each library has its own documentation to explain its architecture.

raulraja (approved):

PR #180, opened 2017-04-04
Fix broken links in README.md

### LINKED ISSUE #179: Broken links in README.md
The links for `Free algebras`, `@free`, and `@modules` are broken. They need ".md" added.

Related to #169 and #175

suhasgaddam:
Fixes #179


diesalbla:
@suhasgaddam When I look at the docs in the present `master`, be it either locally or in the [web page](http://frees.io/), those links are not broken.

Note that, the way the `jekyll` or the `sbt-microsites` work, each file like `algebras.md` and `modules.md` has a permalink definition at the top. Thus, any link in the site that goes to that permalink, will be redirected to the file.

suhasgaddam:
@diesalbla 

It looks like there are 2 `README.md` files.
* https://github.com/47deg/freestyle/blob/master/docs/src/main/tut/README.md
  * I am guessing this is the source for sbt-microsites/jekyll
  * This page seems broken in GitHub's Markdown viewer
* https://github.com/47deg/freestyle/blob/master/README.md
  * This one has broken links through GitHub's Markdown viewer
  * I edited this file at the root of the repo which I don't think it is compiled with sbt-microsites/jekyll
  * Also, this file looks very similar to the other `README.md`

diesalbla:
I see no problem in the `docs` pages looking broken in GitHub's Markdown viewer. They are not for it. 

@raulraja On what @suhasgaddam comments: ¿should the `README.md` at the root folder contain exactly the same as the `docs`? ¿Or should it rather be focused on how to use-integrate-build-test the library itself?

raulraja:
@diesalbla the README in docs/ is the one for the [frees.io](http://frees.io) site and the one at the root is the one for github. Eventually the one in github will have different content and they should not be the same.
The links of both the docs and github should work when navigating inside the docs/ folder whether that happens in github or in the site. That is why all files should be `docs/section/README.md` so github automatically displays the proper docs as niavigating through folders. Currently a few of those are broken because they are not `README` files inside a folder.

raulraja:
What we can do in this PR or another to get this solved is ensure all docs files conform to the following structure: `docs/section/README.md` This will make it work both in github and the doc site provided that permalinks as defined in the Jekyll site looks like `docs/section` without the `README` piece. @suhasgaddam if you want to give it a shot we can get everything fixed for all docs both on github and the frees.io site. 

An example of a currently working link that works both on github and jekyll is : https://github.com/47deg/freestyle/tree/master/docs/src/main/tut/docs/fetch 

If you notice by navigating to that folder Github automatically displays the README file.
Since Jekyll allows us to customize permalinks, we set the permalink in that doc to `/docs/fetch` and jekyll also shows the README file.

If we apply this same technique to the rest of files we will have the best of both worlds.

suhasgaddam:
Directories that need `README.md`
* https://github.com/47deg/freestyle/tree/master/docs/src/main/tut/docs/core
* https://github.com/47deg/freestyle/tree/master/docs/src/main/tut/docs/http
* https://github.com/47deg/freestyle/tree/master/docs/src/main/tut/docs/integrations
* https://github.com/47deg/freestyle/tree/master/docs/src/main/tut/docs/patterns

Since these are parent topics, does it make sense to have a simple `README.md` with links/navigation to the children markdown files?

raulraja:
They all really need to be README including children AFAIK, otherwise navigating to child nodes won't work on github because github expects the files to actually be called README to display them when navigating to a folder view.

suhasgaddam:
Well, I fixed all the links for GitHub which then broke all the links for sbt-microsite/jekyll...

raulraja:
Maybe the permalinks in the menu and each file header needs to be adjusted?

README.md
```diff
@@ -29,7 +29,7 @@ import freestyle._
 
 # Automatic Dependency Injection
 
-Freestyle includes all the implicit machinery necessary to achieve seamless dependency injection of [`@free`](/docs/src/main/tut/docs/core/algebras) and [`@module`](/docs/src/main/tut/docs/core/modules) Algebras.
+Freestyle includes all the implicit machinery necessary to achieve seamless dependency injection of [`@free`](/docs/src/main/tut/docs/core/algebras/) and [`@module`](/docs/src/main/tut/docs/core/modules/) Algebras.
```
raulraja: These links should point to the actual urls at http://frees.io

docs/src/main/tut/README.md
```diff
@@ -37,7 +37,7 @@ import freestyle._
 <div class="row" markdown="1">
 <div class="col-md-6 col-md-push-6" markdown="1">
 # Automatic Dependency Injection
-Freestyle includes all the implicit machinery necessary to achieve seamless dependency injection of [`@free`](/docs/core/algebras) and [`@module`](/docs/core/modules) Algebras.
+Freestyle includes all the implicit machinery necessary to achieve seamless dependency injection of [`@free`](/docs/src/main/tut/docs/core/algebras) and [`@module`](/docs/src/main/tut/docs/core/modules) Algebras.
```
raulraja: These are internal links in frees.io also. To get them to work in github and inside frees.io we may just use the absolute urls in frees.io to whatever content they are pointing to.

raulraja (changes requested):

raulraja (approved):

PR #181, opened 2017-04-05
Add a doc section for freestyle-async

### LINKED ISSUE #148: Docs: Async module

peterneyens:
Resolves #148.

Under which menu section should I put a link to this section?


raulraja (approved):
I think it makes sense under effects. What do you think? We can also split each effect into its own file if it makes more sense since anchor links are still not getting generated for h1 sections.

Issue #177, opened 2017-04-04
Make Macros Hygienic


diesalbla:
The current implementation of the Scala macros `@free` and `@module` uses some names for types, which are only internal to the definition, but with a fixed name. That has the risk of a name collision with the parameters given by the user's definition. The goal of this ticket is to avoid this by making these macros [hygienic](http://docs.scala-lang.org/overviews/quasiquotes/hygiene).

PR #186, opened 2017-04-07
FS-177 Make the `@module` and `@free` macros hygienic

diesalbla:
This PR fulfills issue #177. We 

Intuitively, [macro hygiene](http://docs.scala-lang.org/overviews/quasiquotes/hygiene) refers to ensuring that the names used in the generated code for values, types, or imports, do not collide with those of the context in which the macro is used. In our case, that means: 
- The names of the type parameters used in the definitions of the `Handler` or the `To` should not be     confused with those of the context; but more important, should not shadow (accidentally _capture_) the type parameters given by the user for the `trait`. To achieve this, we use `freshTypename`. 
- The names for the auxiliary members inside the generated companion objects should not collide with those members defined by the user in the trait.  We use `freshTermName` for this. 
- The imports in the generated code should be those referred to in the macro implementation code, and not in the context of the call. We solve this by qualifying imports with `_root_`.   

Along the way, we found a bug in the `@macro` implementation: the type name `F[_]` was hard-coded into the macro, so it was not possible to define a `@macro`-annotated trait whose parameter was, say, `G[_]`. We have repaired that.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,14 +1,12 @@
-# Cats jvmopts
-# see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
+# Cats jvmopts see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
 -Dfile.encoding=UTF8
--Xms1G
-#-Xmx6G
+-Xms2G
 -Xmx2G
--XX:MaxPermSize=512M
--XX:ReservedCodeCacheSize=250M
+-XX:MaxMetaspaceSize=512M
+-XX:ReservedCodeCacheSize=128M
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
 # effectively adds GC to Perm space
+-XX:+UseConcMarkSweepGC
 -XX:+CMSClassUnloadingEnabled
 # must be enabled for CMSClassUnloadingEnabled to work
--XX:+UseConcMarkSweepGC
--- a/freestyle/shared/src/main/scala/freestyle/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/free.scala
@@ -16,49 +16,47 @@
 
 package freestyle
 
-import scala.annotation.{compileTimeOnly, StaticAnnotation}
-import scala.language.experimental.macros
-import scala.reflect.macros.blackbox
+import scala.reflect.macros.blackbox.Context
 
-trait FreeModuleLike {
-  type Op[A]
-}
+object freeImpl {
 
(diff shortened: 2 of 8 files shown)


.jvmopts
```diff
@@ -1,14 +1,12 @@
...
+# Cats jvmopts see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
 -Dfile.encoding=UTF8
--Xms1G
-#-Xmx6G
+-Xms2G
 -Xmx2G
--XX:MaxPermSize=512M
--XX:ReservedCodeCacheSize=250M
+-XX:MaxMetaspaceSize=512M
+-XX:ReservedCodeCacheSize=128M
```
diesalbla, line 6: The CodeCacheSize parameter has to do with the JIT Code Cache. Since most of our code (tests) is only runned once, we can use a smaller one. 

_Edit_: [Here](https://dzone.com/articles/understanding-the-java-memory-model-and-the-garbag) is a reference that may come useful.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -16,49 +16,54 @@
...
-}
+  object messages {
+
+    val invalid = "Invalid use of the `@free` annotation"
 
-object free {
+    val abstractOnly = "The `@free` macro annotation can only be applied to either a trait or an abstract class."
 
-  def impl(c: blackbox.Context)(annottees: c.Expr[Any]*): c.universe.Tree = {
+    val noCompanion = "The trait or class annotated with `@free` must have no companion object."
```
raulraja: This is a limitation that we may want to remove if it's easy to just add the generated code into an existing companion object.

freestyle/shared/src/test/scala/freestyle/Utils.scala
```diff
@@ -27,15 +27,15 @@ object algebras {
   }
 
   @free
-  trait SCtors2[F[_]] {
-    def i(a: Int): FreeS[F, Int]
-    def j(a: Int): FreeS[F, Int]
+  trait SCtors2[G[_]] {
```
raulraja, line 30: Are these to tests what we discussed regarding `F[_]` being hardcoded?

raulraja (approved):
Great cleanup @diesalbla! thanks!

freestyle/shared/src/test/scala/freestyle/Utils.scala
diesalbla: Yes. To have some tests on the matter.

peterneyens (commented):
Certainly a nice clean up , :+1: !

PR #194, opened 2017-04-08
FS-93 - 1 Auto-generate `F[_]` parameter from 

diesalbla:
This PR carries a few steps regarding ticker #93. The goal of that ticket is to remove the need to specify the `F[_]` high-kind parameter in the `@free` and `@module` annotation. This PR carries a partial solution, which helps us to remove the parameter from the `@free`-annotated traits. The approach followed in this PR is: 

- We introduce a trait `EffectLike[F[_]]`, that contains a couple of type aliae `OpSeq[A]` and `OpPar[A]`, for `FreeS[F,A]` and `FreeS.Par[F,A]`, respectively.
- We use macro `freshTypeName` to generate a high-kind type `F` from the `@free` annotated type. This name should not collide with any one chosen by the user. 
- We change the macro to edit the original `@free`-annotated trait, to introduce the extra param `F[_]`, and add its inheritance from `EffectLike[F]`.

### DIFF
--- a/tests/src/test/scala/tests.scala
+++ b/tests/src/test/scala/tests.scala
@@ -27,7 +27,8 @@ class tests extends WordSpec with Matchers {
       import org.ensime.pcplod._
       withMrPlod("pcplodtest.scala") { pc =>
         pc.typeAtPoint('result) shouldBe Some("Option[Int]")
-        pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
+        // Disabled test: the name of the parameter of `FreeS` is no longer known.
+        //pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
         pc.typeAtPoint('handler) shouldBe Some("pcplodtest.PcplodTestAlgebra.Handler")
         pc.messages should be a 'empty
       }
--- a/tests/src/test/scala/tests.scala
+++ b/tests/src/test/scala/tests.scala
@@ -27,8 +27,7 @@ class tests extends WordSpec with Matchers {
       import org.ensime.pcplod._
       withMrPlod("pcplodtest.scala") { pc =>
         pc.typeAtPoint('result) shouldBe Some("Option[Int]")
-        // Disabled test: the name of the parameter of `FreeS` is no longer known.
-        //pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
+        pc.typeAtPoint('test) shouldBe Some("(n: Int)cats.free.FreeApplicative[F,Int]")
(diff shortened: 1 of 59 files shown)


tests/src/test/scala/tests.scala
```diff
@@ -27,7 +27,8 @@ class tests extends WordSpec with Matchers {
       import org.ensime.pcplod._
       withMrPlod("pcplodtest.scala") { pc =>
         pc.typeAtPoint('result) shouldBe Some("Option[Int]")
-        pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
+        // Disabled test: the name of the parameter of `FreeS` is no longer known.
+        //pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
```
fommil: hmm, this looks like it might be a bug in the macro...

diesalbla: Maybe, but the point of the changes is that, the placeholder variable `F[_]`, instead of being part of the user-provided code in the `@free`-annotated `trait`, is now a fresh type name generated by the macro.

diesalbla: If the test is left activated, it fails with an error message such as the following one:
```scala
[info] - should generate code that works in the presentation compiler *** FAILED ***
[info]   Some("(n: Int)cats.free.Free[[β$0$]cats.free.FreeApplicative[F,β$0$],Int]") 
was not equal to Some("(n: Int)freestyle.FreeS[_,Int]") (tests.scala:31)
```
It seems that the aliases for `OpSeq` and `OpPar` are now being expanded. I have made a few small changes to the test, so now we may keep it.

raulraja (approved):
Awesomesauce!

tests/src/test/scala/tests.scala
```diff
@@ -27,7 +27,7 @@ class tests extends WordSpec with Matchers {
       import org.ensime.pcplod._
       withMrPlod("pcplodtest.scala") { pc =>
         pc.typeAtPoint('result) shouldBe Some("Option[Int]")
-        pc.typeAtPoint('test) shouldBe Some("(n: Int)freestyle.FreeS[F,Int]")
+        pc.typeAtPoint('test) shouldBe Some("(n: Int)cats.free.FreeApplicative[F,Int]")
```
diesalbla, line 30: @fommil Does this compromise solution keep with what you were aiming for?

fommil: looks good. But can you please add an example for the common use case of a `Free` algebra?

diesalbla: Ok. I will take care of it on next issue. It may not be easy because when using `FreeS` we are also using a type-name generated by the kind-projector.

PR #195, opened 2017-04-10
Dependency Injection docs

### LINKED ISSUE #132: Docs: Dependency Injection section
We need docs explaining how we automatically add default implicit instances for all @free and @module definitions and how that enables dependency Injection. This section should also use a couple of basic examples on how you can achieve alternative runtime based DI via the Reader `effect`

peterneyens:
Did I miss something, or do we need something else?

I tried to come up with a simple but not too trivial example of using `Reader` / `Kleisli` as target data type where you do not want the environment or config to be part of your algebra (like with `ReaderM`), but I couldn't come up with something small but still useful.

I don't know if it makes sense to add something trivial like : 

```scala
@free trait Calc[F[_]] {
  def sum(a: Int, b: Int): FreeS[F, Int]
}

type IntReader[A] = Reader[Int, A]

implicit val handler: Calc.Handler[IntReader] = new Calc.Handler[IntReader] {
  def sum(a: Int, b: Int): IntReader[Int] = Reader(n => a + (b * n))
}

Calc[Calc.Op].sum(1, 2).exec[IntReader].run(5) // 11 (= 1 + 2 x 5)
```

Resolves #132.

### DIFF
--- a/docs/src/main/resources/microsite/data/menu.yml
+++ b/docs/src/main/resources/microsite/data/menu.yml
@@ -59,7 +59,7 @@ options:
         url: docs/effects/Cache/
 
       - title: Dependency Injection
-        url: TODO.html
+        url: docs/patterns/dependency-injection
 
       - title: Configuration
         url: docs/patterns/config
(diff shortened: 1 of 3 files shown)


raulraja:
I think we need to mention here that we include synthetic implicits in all Algebras and Modules so you can inject them anywhere by requiring them implicitly, that should be the main part of the docs. then provide examples of that. The actual Reader example should be mentioned as an alternative secondary technique just to show that freestyle supports other styles of DI via the `reader` effect where you can inject other things beside config

PR #198, opened 2017-04-15
EitherM effect algebra

raulraja:
This new algebra allows the same semantics as the `ErrorM` effect algebra but allows the user to set `E` to something else beside `Throwable`


juanpedromoreno (approved):
LGTM

Issue #221, opened 2017-04-18
Review module macro


dominv:
This issue is related with #185. When the code snippets from the freestyle docs is run without using objects: 

```scala
import freestyle._
import freestyle.implicits._

  @free trait Validation[F[_]] {
    def minSize(s: String, n: Int): FreeS.Par[F, Boolean]
    def hasNumber(s: String): FreeS.Par[F, Boolean]
  }

  @free trait Interaction[F[_]] {
    def tell(msg: String): FreeS[F, Unit]
    def ask(prompt: String): FreeS[F, String]
  }
  
  @module trait Application[F[_]] {
    val validation: Validation[F]
    val interaction: Interaction[F]
  }
```

the following error is returned: 
```
exception during macro expansion: 
scala.ScalaReflectionException: <none> is not a class
	at scala.reflect.api.Symbols$SymbolApi.asClass(Symbols.scala:272)
	at scala.reflect.api.Symbols$SymbolApi.asClass$(Symbols.scala:272)
	at scala.reflect.internal.Symbols$SymbolContextApiImpl.asClass(Symbols.scala:88)
	at freestyle.openUnion$.materializeImpl(module.scala:92)
```
The line 92 in `module.scala` is:
```scala
    val algebras   = findAlgebras(weakTypeOf[A].typeSymbol.companion.asClass)
```
And its failing when it calls to the companion method:
```
weakTypeOf[A]: Application.type
weakTypeOf[A].typeSymbol: object Application
weakTypeOf[A].typeSymbol.companion: <none>
```

On the other hand, when using objects, like: 

```scala
import freestyle._
import freestyle.implicits._

object algebras {
  @free trait Validation[F[_]] {
    def minSize(s: String, n: Int): FreeS.Par[F, Boolean]
    def hasNumber(s: String): FreeS.Par[F, Boolean]
  }

  @free trait Interaction[F[_]] {
    def tell(msg: String): FreeS[F, Unit]
    def ask(prompt: String): FreeS[F, String]
  }
}

object modules {
  import algebras._
  
  @module trait Application[F[_]] {
    val validation: Validation[F]
    val interaction: Interaction[F]
  }
}

the results are:
```
```
weakTypeOf[A]: Evaluator__3b8302c6c45f12ee4172cc7128601deeeff5e5d0_1123939886.modules.Application.type
weakTypeOf[A].typeSymbol: object Application
weakTypeOf[A].typeSymbol.companion: trait Application
```

dominv:
[...]

PR #230, opened 2017-04-18
FS-221 Remove wrapper objects in tests

diesalbla:
This ticket tries to address the possible problems mentioned in ticket #221. This issue was whether it is possible to define `@free` and `@module` without having to wrap these in an `object`, which was failing at a Kazari instance. In this PR, we change the tests of the core module so as to use an organization without those wrapper objects. We also add some comments for the code of the `@module` macro.


raulraja (approved):
Cool!

Issue #229, opened 2017-04-18
Cache docs and potentially others not using tut.


raulraja:
Some of the docs have simple `Scala` blocks instead of `tut:book` and are not getting compiled.

diesalbla:
We have already passed to `tut:book` those code fragments that are intended to be executable. I have left as `Scala` those that describe types or laws.

PR #233, opened 2017-04-18
FS-229 Pass codeblocks to tut

diesalbla:
This PR fulfills #229.

We pass some of the blocks of code in the documentation to the `tut`
format. We leave as  `Scala` those that are not to be executed.


raulraja (approved):

raulraja:
Looks great!, need to resolve conflicts before this can be merged.

Issue #234, opened 2017-04-19
Attempt to Remove `OpSeq` and `OpPar` in favor of a single `FS`


raulraja:
@diesalbla and I discussed offline that it may be possible to treat Ops as Par exclusively and provide proper lifting in to monads implicitly where appropriate. This issue tracks the attempt to remove the verbosity while preserving the feature of paralleling any applicative style composition of those new `FS` ops. This does not need to go in on 0.1.0 but if we have a solution before release we will consider it for inclusion.

diesalbla:
@raulraja There is one possible further improvement that we may now consider: we could avoid the `FS[_]` wrapper completely. What we would do is that, in a `@free` annotated class or method, any abstract method declaration `def ann(x: Int): Bob` would be understood as being an atomic operation of type `FS[Bob]`, i.e., `FreeS.Par[F$, Bob]`, where   `F$[_]` is the type parameter inserted by the macro.

raulraja:
I think we need the wrapper so we can express at some point that it is a monad/applicative and users can directly add concrete methods that compose those return types. I think removing that will add confusion when you realize that you can compose monadically two different types such as `Bob` and `Jane`.

PR #237, opened 2017-04-19
FS-234 Unify OpPar and OpSeq

diesalbla:
This PR addresses issue #234. The goal of this issue is that, in the effect-algebra traits (those annotated with `@free`), there should be no separation between _parallel_ or _sequential_ operations. To achieve this, the changes in this PR modify the `@free` macro annotation as follows: 

* In the `EffectLike` trait we add a new type alias, `FS[A] = FreeS.Par[A]`. 
* We add a sanity check: in a `@free`-annotated trait, all abstract methods should have as return type `FS[_]`. 
* We keep the aliases `OpPar` and `OpSeq`, to be used for those methods in the trait that are declared and implemented, since this implementation is based on the monadic or applicative operations for `FreeS` and `FreeS.Par`. 
* (not directly related to the issue) we modify the `FreeS.inject` method, so that we build the _Injector_ object once, using the base `Inject` instance.

### DIFF
--- a/docs/src/main/tut/docs/core/interpreters/README.md
+++ b/docs/src/main/tut/docs/core/interpreters/README.md
@@ -24,12 +24,12 @@ import freestyle._
 import cats.implicits._
 
 @free trait KVStore {
-  def put[A](key: String, value: A): OpSeq[Unit]
-  def get[A](key: String): OpSeq[Option[A]]
-  def delete(key: String): OpSeq[Unit]
+  def put[A](key: String, value: A): FS[Unit]
+  def get[A](key: String): FS[Option[A]]
+  def delete(key: String): FS[Unit]
   def update[A](key: String, f: A => A): OpSeq[Unit] =
-    get[A](key) flatMap {
-      case Some(a) => put[A](key, f(a))
+    get[A](key).freeS flatMap {
+      case Some(a) => put[A](key, f(a)).freeS
       case None => ().pure[OpSeq]
     }
 }
@@ -85,8 +85,8 @@ To illustrate interpreter composition, let's define a new algebra `Log` which we
 
 ```tut:book
 @free trait Log {
-  def info(msg: String): OpSeq[Unit]
-  def warn(msg: String): OpSeq[Unit]
+  def info(msg: String): FS[Unit]
+  def warn(msg: String): FS[Unit]
 }
 ```
 
--- a/docs/src/main/tut/docs/core/interpreters/README.md
+++ b/docs/src/main/tut/docs/core/interpreters/README.md
@@ -27,10 +27,10 @@ import cats.implicits._
   def put[A](key: String, value: A): FS[Unit]
   def get[A](key: String): FS[Option[A]]
   def delete(key: String): FS[Unit]
-  def update[A](key: String, f: A => A): OpSeq[Unit] =
+  def update[A](key: String, f: A => A): FS.Seq[Unit] =
     get[A](key).freeS flatMap {
       case Some(a) => put[A](key, f(a)).freeS
-      case None => ().pure[OpSeq]
+      case None => ().pure[FS.Seq]
     }
 }
 ```
--- a/freestyle/shared/src/main/scala/freestyle/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/free.scala
@@ -34,6 +34,7 @@ object freeImpl {
     val MM = freshTypeName("MM$") // MM Target of the Handler's natural Transformation
     val LL = freshTypeName("LL$") // LL is the target of the Lifter's Injection
     val AA = freshTypeName("AA$") // AA is the parameter inside type applications
+    val inj = freshTermName("toInj")
 
     def fail(msg: String) = c.abort(c.enclosingPosition, msg)
 
@@ -41,6 +42,7 @@ object freeImpl {
     val invalid = "Invalid use of the `@free` annotation"
     val abstractOnly = "The `@free` annotation can only be applied to a trait or to an abstract class."
     val noCompanion = "The trait (or class) annotated with `@free` must have no companion object."
+    val onlyReqs = "In a `@free`-annotated trait (or class), all abstract method declarations should be of type OpSeq[_] or OpPar[_]"
 
     def gen(): Tree = annottees match {
       case Expr(cls: ClassDef) :: Nil =>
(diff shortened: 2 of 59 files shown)


docs/src/main/tut/docs/core/interpreters/README.md
```diff
@@ -24,12 +24,12 @@ import freestyle._
...
-  def put[A](key: String, value: A): OpSeq[Unit]
-  def get[A](key: String): OpSeq[Option[A]]
-  def delete(key: String): OpSeq[Unit]
+  def put[A](key: String, value: A): FS[Unit]
+  def get[A](key: String): FS[Option[A]]
+  def delete(key: String): FS[Unit]
   def update[A](key: String, f: A => A): OpSeq[Unit] =
-    get[A](key) flatMap {
-      case Some(a) => put[A](key, f(a))
+    get[A](key).freeS flatMap {
```
diesalbla: This is the one point in which unifying both types adds a bit of chrome: now we need to manually raise the `FS[A]` to a `FreeS[ $F, A]`. The same happens with the result of `put` in the next line.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -117,30 +109,23 @@ object freeImpl {
...
-        val (liftType, impl) = tpt match {
-          case tq"OpSeq[$aa]" => (tpt, q"FreeS.liftPar($injected)" )
-          case tq"OpPar[$aa]" => (tpt, injected )
-          case _ => // Note: due to filter in getRequestDefs, this case is unreachable.
-            fail(s"unknown abstract type found in @free container: $tpt : raw: ${showRaw(tpt)}")
-        }
+      def raiser: DefDef =
+        q"""
+          override def ${reqDef.name}[..$tparams](...${reqDef.vparamss}): ${reqDef.tpt} =
+            $inj( $ReqC[..${tparams.map(_.name)} ](..${params.map(_.name)}) )
```
diesalbla, line 117: Note that, because `$inj` is appearing in the scope of the value parameters of the method, we need a fresh name; otherwise it may accidentally collide with any of those in `reqDef.vparamasss`.

freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -41,7 +41,7 @@ class implicitsTests extends WordSpec with Matchers {
     "enable sequence" in {
       implicit val optionHandler = interps.optionHandler1
       val s = SCtors1[SCtors1.Op]
-      val program = List(s.x(1), s.x(2)).sequence[FreeS[SCtors1.Op, ?], Int]
+      val program = List(s.x(1), s.x(2)).sequence[FreeS.Par[SCtors1.Op, ?], Int]
```
diesalbla: Another point in which, by being now parallel by default, we have to make a special change.

freestyle/shared/src/test/scala/freestyle/implicits.scala
```diff
@@ -53,7 +53,7 @@ class implicitsTests extends WordSpec with Matchers {
 
     "provide a custom implicit not found message for a missing Capture instance" in {
       shapeless.test.illTyped(
-        """SCtors1[SCtors1.Op].x(1).exec[scala.util.Try]""",
+        """SCtors1[SCtors1.Op].x(1).freeS.exec[scala.util.Try]""",
```
diesalbla: Another case in which we need to lift the value to the squential mode.

http/play/src/test/scala/PlayTests.scala
```diff
@@ -58,7 +58,7 @@ class PlayTests extends AsyncWordSpec with Matchers {
 
     "FreeS programs can interact with a given request and used as returned values in Play actions" in {
       Action.async { request =>
-        Noop[Noop.Op].noop.map(_ => Results.Ok(request.method))
+        Noop[Noop.Op].noop.map(_ => Results.Ok(request.method)).freeS
```
diesalbla: Another case in which we need to raise the result to sequential.

freestyle/shared/src/main/scala/freestyle/package.scala
```diff
@@ -58,15 +58,14 @@ package object freestyle {
     def liftPar[F[_], A](freeap: FreeS.Par[F, A]): FreeS[F, A] =
       Free.liftF(freeap)
 
-    def inject[F[_], G[_]]: FreeSParInjectPartiallyApplied[F, G] =
+    def inject[F[_], G[_]](implicit I: Inject[F, G]): FreeSParInjectPartiallyApplied[F, G] =
```
peterneyens: If you pass the implicit `Inject` instance here, I don't think we need `FreeSParInjectPartiallyApplied`.

Instead it could probably be:
```scala
def inject[F[_], G[_]](implicit I: Inject[F, G]): F ~> FreeS.Par[G, ?] =
  λ[F ~> FreeS.Par[G, ?]](fa => FreeApplicative.lift(I.inj(fa)))
```

freestyle/shared/src/test/scala/freestyle/implicits.scala
peterneyens: Just wondering, why do we need to lift to `FreeS`?
You can also call `exec` on `FreeS.Par` right?

Or am I missing what this test is about?

peterneyens: The test messages (`"provide a custom ...`) of this test and the previous one also seem wrong, I think they need to be swapped to make sense.

Maybe we can fix this in this PR?.

http/play/src/test/scala/PlayTests.scala
raulraja: We can fix this with another implicit conversion in the play module if you are in the context of `FreeS.Par`

freestyle/shared/src/test/scala/freestyle/implicits.scala
diesalbla: The test is about the conversion from `FreeS.Par[F, _]` or `FreeS[F, _]` to `Future` (which is what the `play` integration provides). I have added another _conversion_ to the main code, and a test for it.

freestyle/shared/src/main/scala/freestyle/package.scala
```diff
@@ -111,8 +105,8 @@ package object freestyle {
      */
     def freeS: FreeS[F, A] = FreeS.liftPar(fa)
 
-    def exec[G[_]: Applicative](implicit interpreter: F ~> G): G[A] =
```
diesalbla, line 114: This change was used to avoid a strange error in the test for implicit resolution messages.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -19,6 +19,7 @@ package freestyle
 import scala.reflect.macros.blackbox.Context
 
 trait EffectLike[F[_]] {
+  final type FS[A] = FreeS.Par[F, A]
   final type OpSeq[A] = FreeS[F, A]
```
raulraja: Can these be removed now that there is a single `FS`?

docs/src/main/tut/docs/core/interpreters/README.md
raulraja: If we are keeping `OpSeq` and `OpPar` around for special cases like this one I think we should rename then to something like `FS.Seq` and `FS.Par` so they remain consistent with the naming conventions for return types in `@free` declarations

raulraja (approved):
Outstanding work!

Issue #165, opened 2017-04-01
FQN names should use _root_.


fommil:
this file does not compile https://github.com/fommil/drone-dynamic-agents/blob/0d9cf19fd1fd7804e1fc03477251bcf8afbd5c55/src/main/scala/interpreters/oauth2.scala

mostly because when in the package `interpreters.oauth2.client` and referring to `interpreters.oauth2` one may expect the compiler to search from the root, but it starts looking in `interpreters.oauth2.client.interpreters` and fails to find anything.

Admittedly, my packaging convention here is rubbish and the immediate fix is for me to reorganise in a sensible way. But this does point to a more general bug in the macro.

diesalbla:
The following code hits the same problem of unqualified names in macro and ambiguity in package resolution.
```scala
package issue165
import freestyle._
object issue165
@free trait India { def kolkata: FS[Int] }
@module trait Asia[F[_]] {  val india: India[F] }
```

raulraja:
@fommil This PR https://github.com/47deg/freestyle/pull/239 solves the issue but it introduces other problems that may be more important to most users. Among other things FQN in the macro for those types prevents from using expressions in `@module` annotations in the REPL, tut and other tools that use REPL sessions for compiling snippets of code. We are gonna keep this issue open and the branch around in case we can come up with a better solution that does not have this side effects. Here is a sample REPL session snippet. Any ideas to overcome this issue are welcome. Thanks!

```
[info] Set current project to freestyle (in build file:/home/raulraja/Documents/workspace/diegotest/freestyle/)
> console
[info] Starting scala interpreter...
[info] 
Welcome to Scala 2.12.1 (OpenJDK 64-Bit Server VM, Java 1.8.0_111).
Type in expressions for evaluation. Or try :help.
scala> import freestyle._
import freestyle._
scala> import freestyle.implicits._
import freestyle.implicits._
scala> @free trait X { def x: FS[Int] }
defined trait X
defined object X
scala> @free trait Y { def y: FS[Int] }
defined trait Y
defined object Y
scala> @module trait Z[F[_]] { val x : X[F]; val y : Y[F] }
<macro>:1: error: object Y is not a member of package <root>
type CC$2[AA$2] = Coproduct[_root_.Y.Op, _root_.X.Op, AA$2]
                                   ^
<macro>:1: error: <error> takes no type parameters, expected: one
type CC$2[AA$2] = Coproduct[_root_.Y.Op, _root_.X.Op, AA$2]
[...]

PR #239, opened 2017-04-20
FS-165 Module looks from Root

diesalbla:
This PR addresses ticket #165. It modifies the `@module` macro, so that now it looks for the `Op` types of the algebra using the `_root_ prefix. 

One side effect of this changes is that it restricts the  `@module` macro to use `@free` or `@module` traits with a defined `package`, which excludes those defined in a Scala console session, or (for extension) in a [tut](https://github.com/tpolecat/tut) book. For instance, pasting this to the console: 
```scala
import freestyle._
@free trait Foo { def x: FS[Int] }
@module trait Bar[F] {val y: Foo[F] }
```
would no longer work. For this reason we have to adapt the many `tut:book` items in the `docs`, adding a few Scala files.

### DIFF
--- a/docs/src/main/tut/docs/README.md
+++ b/docs/src/main/tut/docs/README.md
@@ -40,7 +40,6 @@ In the example below we will define two algebras with intermixed sequential and 
 
 ```tut:book
 import freestyle._
-import freestyle.implicits._
 
 @free trait Validation {
   def minSize(s: String, n: Int): FS[Boolean]
@@ -60,6 +59,10 @@ Learn more about [algebras](./core/algebras) in the extended documentation.
 Freestyle algebras can be combined into `@module` definitions which provide aggregation and unification over the
 parametrization of Free programs.
 
+```tut:reset:silent
+import freestyle._
+import freestyle.docs._
+```
 ```tut:book
 @module trait Application[F[_]] {
   val validation: Validation[F]
@@ -80,9 +83,11 @@ Abstract definitions it's all it takes to start building programs that support s
 The example below combines both algebras to produce a more complex program
 
 ```tut:book
+import cats.syntax.cartesian._
+import freestyle.implicits._
+
 def program[F[_]](implicit A: Application[F]) = {
   import A._
-  import cats.implicits._
 
   for {
     userInput <- interaction.ask("Give me something with at least 3 chars and a number on it")
@@ -119,7 +124,7 @@ The mere fact that you provide implicit evidences for the individual steps enabl
 At this point we can run our pure programs at the edge of the world.
 
 ```tut:book
-import cats.implicits._
+import cats.instances.future._
 import scala.concurrent.duration.Duration
 import scala.concurrent.Await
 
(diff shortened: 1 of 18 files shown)


docs/src/main/tut/docs/patterns/logging/README.md
```diff
@@ -17,7 +17,7 @@ libraryDependencies += "com.47deg" %% "freestyle-logging" % "0.1.0"
 
 ### Operations
 
-The set of abstract operations of the `Logging` algebra are specified as follows:
+The set of abstract operations of the [`Logging` algebra](../../../../../../../freestyle-logging/shared/src/main/scala/logging.scala) are specified as follows:
```
peterneyens, line 20: What happens if you click on this link on the microsite?

peterneyens:
Hmm, not being able to create a `module` in the REPL, tut (and I expect scastie), seems kind of a hard price to pay?

docs/src/main/tut/docs/README.md
```diff
@@ -80,9 +83,11 @@ Abstract definitions it's all it takes to start building programs that support s
 The example below combines both algebras to produce a more complex program
 
 ```tut:book
+import cats.syntax.cartesian._
```
peterneyens, line 86: The cats documentation choose to always use `import cats.implicits._` instead of the piecemeal imports as it is more beginner friendly, should we do the same?

diesalbla:
You can create a module, and the changes in the `tut:book` are still creating modules. What it can not do is to use the `@module` or `@free` traits defined in the same console, `tut`, or `scastie`.

docs/src/main/tut/docs/patterns/logging/README.md
diesalbla: Run locally, it fails. It is intended to be use in github, as a link to the source code.

peterneyens:
Creating a `@module` or `@free` traits in the REPL without being able to use them, is not very useful, so my wording may have been wrong, but this still means that it becomes impossible to use freestyle in the REPL which to me seems sort of a big deal.

docs/src/main/tut/docs/README.md
diesalbla: IMHO, it is more beginner-friendly to narrow down, as much as possible, what is it that the code snippet is using from this big `cats` library. Not everybody may know all of `cats`.

peterneyens: Yeah, that was my initial point of view as well, but it is even more difficult to figure out where exactly which syntax is coming from before you know the relation between all the concepts Cats provides, see https://github.com/typelevel/cats/issues/1026.

docs/src/main/tut/docs/patterns/logging/README.md
peterneyens: If we used an absolute link (eg `https://github.com/47deg/freestyle/blob/master/freestyle-logging/shared/src/main/scala/logging.scala`) it would work on the website as well, right?

raulraja (changes requested):
We are putting this on hold because fixing the FQN names creates the side effect of `@module` not working in REPL environments. See comments on #165 for details

diesalbla:
This has fallen far behind, and the problem may not be a major one.

raulraja:
@diesalbla feel free to reopen if you consider this can be fixed now

PR #242, opened 2017-04-25
Tagless Final encoding for Freestyle

### LINKED ISSUE #53: Research generation of other encodings simultaneously for max flexibility
Tagless for ex.

raulraja:
Fixes #53 

This PR brings an alternative to `@free` which instead of generating Free boilerplate conform to the Tagless Final encoding http://okmij.org/ftp/tagless-final/

`@tagless` final algebras conform to the same syntax as `@free` ones both in definition and handlers implementation for easy user adoption:

```scala
@tagless trait TG1 {
  def x(a: Int): FS[Int]
  def y(a: Int): FS[Int]
}

implicit val optionHandler1: TG1.Handler[Option] = new TG1.Handler[Option] {
    def x(a: Int): Option[Int] = Some(a)
    def y(a: Int): Option[Int] = Some(a)
}
```

Tagless final algebras is more performant than @free in the context that it requires fewer allocations and the program is not reified as an in memory data structure. The disadvantage is that they are not stack safe unless the target Monad you are interpreting too is also stack safe.
Freestyle provides two alternatives to achieve stack safety.

1. Interpret to the context of `FreeS[M, ?]` to achieve stack safety thanks to `cats.data.Free` built in reification of operations.

```scala
implicit val stackSafeOptionHandler1: TG1.Handler[FreeS[Option, ?]] = new TG1.Handler[FreeS[Option, ?]] {
    def x(a: Int): FreeS[Option, Int] = FreeS.liftFA(Some(a))
    def y(a: Int): FreeS[Option, Int] = FreeS.liftFA(Some(a))
  }
```

2. Use a `StackSafe` version of the Algebra that delegates to it's `@free` representation and that allows interleaving in `@module` and `@free` programs in places where you require stack safety or you need it to interleave with an already `@free` based stack.

```scala
@module trait App[F[_]] {
    val f1: F1[F]
    val tg1: TG1.StackSafe[F]
    val tg2: TG2.StackSafe[F]
 }

def program[F[_] : F1: TG1.StackSafe : TG2.StackSafe]: FreeS[F, Int] = {
    val tg1 = TG1.StackSafe[F]
    val tg2 = TG2.StackSafe[F]
    val fs = F1[F]
    for {
          a <- fs.a(1)
          tg2a <- tg2.x2(1)
          b <- fs.b(1)
          x <- tg1.x(1)
          tg2b <- tg2.y2(1)
          y <- tg1.y(1)
     } yield a + b + x + y + tg2a + tg2b
}
```


freestyle-tagless/shared/src/test/scala/taglessTests.scala
```diff
@@ -0,0 +1,162 @@
...
+
+    "remain stack safe when interpreted to stack safe monads" in {
+      SOProgram[FreeS[Option, ?]](0).exec[Option] shouldBe Option(iterations)
+    }
+
+  }
+
+}
+
+object algebras {
```
diesalbla, line 73: ¿Do they work without being wrapped in `object`?

freestyle-tagless/shared/src/main/scala/tagless.scala
```diff
@@ -0,0 +1,146 @@
...
+      val stackSafeHandler = freshTermName("stackSafeHandler$")
+
+      q"""
+        object ${Eff.toTermName} {
+
+          trait Handler[$MM[_], ..$TTs] extends $Eff[$MM, ..$tns] {
+            ..${requests.map(_.handlerDef)}
+          }
+
+          @_root_.freestyle.free trait StackSafe {
```
diesalbla, line 128: I wonder if, instead of having here the `@free` annotation, we could just extract the `StackSafe` part to another expression and make a call to the `freestyle.freeImpl.free`.

freestyle-tagless/shared/src/test/scala/taglessTests.scala
```diff
@@ -0,0 +1,163 @@
...
+          y <- tg1.y(1)
+        } yield a + b + x + y + tg2a + tg2b
+        x
+      }
+      import TG1._
+      program[App.Op].exec[Option] shouldBe Option(6)
+    }
+
+    "blow up the stack when interpreted to stack unsafe monads" in {
+      assertThrows[StackOverflowError] {
```
peterneyens: It seems this throws a `scala.scalajs.js.JavaScriptException` in ScalaJS instead of `StackOverFlowError`.

freestyle-tagless/shared/src/main/scala/tagless.scala
raulraja: The problem isthat `free` is a macro annotation and it requirees `annotees` as argument and internally validates the structure as being annotated. Furthermore all of it's functions are nested inner functions not visible from the outside and moving those out will increase verbosity in context getting passed around. Not sure it's worth it to refactor all macros to make this an internal call

freestyle-tagless/shared/src/test/scala/taglessTests.scala
raulraja: Yep, just organized them like this to reuse them in tests but before they were inside the test blocks themselvess

raulraja:
@diesalbla @peterneyens addressed all your comments, let me know if this is good to merge once it clears Travis.

diesalbla (approved):

Issue #238, opened 2017-04-20
Remove F[_] parameter from `@module` annotation


diesalbla:
Having already merged PR #194, which removed the `F[_]` parameter from the `@free` annotation as intended by #93, we now have to try to remove it (generate it automatically) from the `@module` annotation as well.

PR #243, opened 2017-04-25
FS-238 Remove F from Module

diesalbla:
This PR fulfills ticket #238. 

We refactor the `@module` macro so as to avoid the use of the `F[_]` parameter, which is now a fresh type name,  added by the macro to the trait and vals.

As a smaller pass-by optimisations, we change the inheritance of the market trait `ModuleLike` out of the object companion and add it to the trait, so as to make the recursive test simpler.

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/module.scala
+++ b/freestyle/shared/src/main/scala/freestyle/module.scala
@@ -16,7 +16,6 @@
 
 package freestyle
 
-import scala.reflect.internal._
 import scala.reflect.macros.whitebox.Context
 
 trait FreeModuleLike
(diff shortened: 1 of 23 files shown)


freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -52,9 +50,7 @@ object openUnion {
       // cs is a `@module` is the _class_ of the `object` extends from `FreeModuleLike`
       // Note that a trait's companion object has a _real_ class, unknown, to rule its behaviour.
       def isModuleFS(cs: ClassSymbol): Boolean =
-        cs.companion.asModule.moduleClass.asClass.baseClasses.exists {
-          case x: ClassSymbol => x.name == TypeName("FreeModuleLike")
-        }
+        cs.baseClasses.exists( _.name == TypeName("FreeModuleLike") )
```
peterneyens, line 44: Should we update the comment above?

diesalbla: Yes. The `cs` is the `ClassSymbol` that points to the `@module`-annotated trait. What the remove line was doing was "go to the companion object and see if its _class_ inherits from `FreeModuleLike`". Note that _its class_ is the one that defines the behaviour and operation of the object, but it is not the one that the companion object is supposed to accompany.

raulraja (approved):
Excellent work @diesalbla !

PR #246, opened 2017-04-26
Fix compiler errors Finch Todo example

peterneyens:
To use the finch integration that makes it possible to use a `FreeS` or `FreeS.Par` in somthing like `get(int) {  ... }`, there needs to be a `FooHandler[Id]` or a `FooHandler[Future]` in scope.

There is also nothing in Freestyle which automatically composes implicitly available natural transformations, so I called `andThen` explicitly in `todo.runtime.implicits`.


raulraja (approved):
Thanks @peterneyens, Do you think we should add something to Freestyle to compose implicit FunctionKs automatically?

PR #250, opened 2017-04-27
Simplify Exec

diesalbla:
We unify the `exec` syntactic methods for `FreeS` and `FreeS.Par`, so that now all `exec` methods take an `FSHandler`. We also replace in a few places the use of `FunctionK` or `~>` with `FSHandler.

### DIFF
--- a/freestyle-cache/shared/src/test/scala/TestContext.scala
+++ b/freestyle-cache/shared/src/test/scala/TestContext.scala
@@ -30,12 +30,12 @@ trait CacheTestContext extends BeforeAndAfterEach { self: Suite =>
   private[this] implicit val rawMap: KeyValueMap[Id, String, Int] =
     new ConcurrentHashMapWrapper[Id, String, Int]
 
-  private[this] implicit val idInt: FSHandler[Id, Id] = FunctionK.id[Id]
+  private[this] implicit val idHandler: FSHandler[Id, Id] = FunctionK.id[Id]
 
   protected[this] final val provider = new KeyValueProvider[String, Int]
 
   protected[this] implicit val interpret: provider.CacheM.Handler[Id] =
-    provider.implicits.cacheHandler(rawMap, idInt)
+    provider.implicits.cacheHandler(rawMap, idHandler)
 
   override def beforeEach = rawMap.clear
 
--- a/freestyle-cache/shared/src/test/scala/TestContext.scala
+++ b/freestyle-cache/shared/src/test/scala/TestContext.scala
@@ -18,6 +18,7 @@ package freestyle.cache
 
 import cats.arrow.FunctionK
 import cats.{~>, Applicative, Id}
+import freestyle.FSHandler
 import freestyle.cache.hashmap._
 import org.scalatest.{BeforeAndAfterEach, Suite}
 
@@ -29,7 +30,7 @@ trait CacheTestContext extends BeforeAndAfterEach { self: Suite =>
   private[this] implicit val rawMap: KeyValueMap[Id, String, Int] =
     new ConcurrentHashMapWrapper[Id, String, Int]
 
-  private[this] implicit val idInt: Id ~> Id = FunctionK.id[Id]
+  private[this] implicit val idInt: FSHandler[Id, Id] = FunctionK.id[Id]
 
   protected[this] final val provider = new KeyValueProvider[String, Int]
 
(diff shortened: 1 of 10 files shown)


http/http4s/src/main/scala/http/http4s.scala
```diff
@@ -17,17 +17,20 @@
...
+      implicit NT: FSHandler[F, G],
       G: Monad[G],
       EE: EntityEncoder[G[A]]): EntityEncoder[FreeS[F, A]] =
     EE.contramap((f: FreeS[F, A]) => f.exec[G])
 
+  implicit def freeSParEntityEncoder[F[_], G[_], A](
+      implicit NT: FSHandler[F, G],
+      G: Applicative[G],
+      EE: EntityEncoder[G[A]]): EntityEncoder[FreeS.Par[F, A]] =
+    EE.contramap((f: FreeS.Par[F, A]) => f.exec[G])
```
peterneyens, line 35: :+1: This previously worked if `G` was a `Monad` as well through the implicit `FreeS.Par` to `FreeS` conversion, but I didn't think of the case where you interpret to "just" an `Applicative`.

So thanks for cleaning up after me!

http/play/src/main/scala/play.scala
```diff
@@ -17,24 +17,20 @@
...
 
 package play {
 
   object implicits {
 
     implicit def seqToFuture[F[_], A](prog: FreeS[F, A])(
-        implicit I: ParInterpreter[F, Future],
+        implicit I: FSHandler[F, Future],
         EC: ExecutionContext
     ): Future[A] = prog.exec[Future]
```
peterneyens: Should we also add an implicit for the case where you have a `ParInterpreter` in scope?

freestyle-doobie/src/main/scala/doobie.scala
```diff
@@ -16,7 +16,7 @@
 
 package freestyle
 
-import cats.{~>, Foldable}
+import cats.Foldable
```
peterneyens: It seems we aren't using `Foldable` in this file, so we can remove this as well.

freestyle-cache/shared/src/test/scala/TestContext.scala
```diff
@@ -29,7 +30,7 @@ trait CacheTestContext extends BeforeAndAfterEach { self: Suite =>
   private[this] implicit val rawMap: KeyValueMap[Id, String, Int] =
     new ConcurrentHashMapWrapper[Id, String, Int]
 
-  private[this] implicit val idInt: Id ~> Id = FunctionK.id[Id]
+  private[this] implicit val idInt: FSHandler[Id, Id] = FunctionK.id[Id]
```
peterneyens: Just a very small minor thing, but can we rename `idInt` to `idHandler` while we are changing things here?
(I find using `Int` as a shorthand for interpreter a bit confusing)

freestyle/shared/src/test/scala/freestyle/Utils.scala
```diff
@@ -99,42 +97,42 @@ trait StateProp {
 
 object interps {
 
-  implicit val optionHandler1: FunctionK[SCtors1.Op, Option] = new SCtors1.Handler[Option] {
+  implicit val optionHandler1: FSHandler[SCtors1.Op, Option] = new SCtors1.Handler[Option] {
```
peterneyens: Just wondering, do we prefer `FSHandler[SCtors1.Op, Option]` over `SCtors1.Handler[Option]` here as the type?

peterneyens:
This doesn't apply directly to this PR, but since we are talking about simplifying `exec`, I am wondering if that is not the wrong name at all.

It doesn't actually execute something, it just translates your program to the target type using a handler. Whether or not something is actually executed, depends on the target type. 

Should we try to come up with a better name. Should I maybe open a new issue ?

raulraja (approved):

http/play/src/main/scala/play.scala
raulraja: yes, we just had that issue today when migrating scala exercises now that all `FS` actions are by default `Par`

raulraja:
@peterneyens let's open an issue, it's true that is not executing for some `M[_]`s such as `Kleisli` where you defer execution to actually applying the target. On the other hand for some commonly used types such as `Future` they will run as soon as you interpret. It will be confusing either way so we may have to properly think what a good name for this is.

peterneyens:
I opened issue #251 to track the possible rename of `exec`.

For this PR, feel free to merge it after the Travis builds go green :+1:

PR #257, opened 2017-04-28
Add debug import to print macro generated trees

andyscott:
This adds a debug implicit that can be imported to configure the macros to print the generated trees to the console. This pattern can be used to add additional debug options.

This was ported over from Iota, which has several debug flags: https://github.com/47deg/iota/blob/v0.0.1/modules/core/src/main/scala/iota/debug/package.scala

The show trees output behaves roughly like this (demo output from Iota):
```
import iota.debug.options.ShowTrees
// import iota.debug.options.ShowTrees

CopK.FunctionK.of[Algebra, Future](evalOrderOp, evalPriceOp, evalUserOp)
// <console>:30: generated tree:
// {
//   final class $anon extends _root_.iota.CopKFunctionK[Algebra, Future] {
//     private[this] val arr0 = evalUserOp.asInstanceOf[_root_.cats.arrow.FunctionK[Any, Future]];
//     private[this] val arr1 = evalOrderOp.asInstanceOf[_root_.cats.arrow.FunctionK[Any, Future]];
//     private[this] val arr2 = evalPriceOp.asInstanceOf[_root_.cats.arrow.FunctionK[Any, Future]];
//     override def apply[A](ca: Algebra[A]): Future[A] = (ca.index: @_root_.scala.annotation.switch) match {
//       case 0 => arr0(ca.value)
//       case 1 => arr1(ca.value)
//       case 2 => arr2(ca.value)
//       case (i @ _) => throw new _root_.java.lang.Exception(StringContext("iota internal error: index ").s().+(i).+(" out of bounds for ").+(this))
//     };
//     override def toString: String = "CopKFunctio...
// res28: iota.CopKFunctionK[Algebra,scala.concurrent.Future] = CopKFunctionK[Algebra, Future]<<generated>>
```

Freestyle Usage:

```
import debug.options.ShowTrees
@free Foo { /* ... */ }
```

Note: If you're using this in test code, the implicit doesn't seem to be picked up correctly if it's nested all the way in the test.

This does not work:
```
"allow smart constructors with no args" in {
  import debug.options.ShowTrees
  @free
  trait NoArgs {
    def x: FS[Int]
  }
}
```

But this does:
```
import debug.options.ShowTrees
"allow smart constructors with no args" in {
  @free
  trait NoArgs {
    def x: FS[Int]
  }
}
```

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/debug/package.scala
+++ b/freestyle/shared/src/main/scala/freestyle/debug/package.scala
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2017 47 Degrees, LLC. <http://www.47deg.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package freestyle
+
+/** Implicit options to configure/control Freestyle's macros
+  */
+package object debug {
+  object optionTypes {
+    sealed trait ShowTrees
+  }
+  import optionTypes._
+
+  object options {
+
+    /** Import this value to have Freestyle print the macro generated code
+      * to the console during compilation
+      */
+    implicit val ShowTrees: ShowTrees = null.asInstanceOf[ShowTrees]
+  }
+}
--- a/freestyle/shared/src/main/scala/freestyle/module.scala
+++ b/freestyle/shared/src/main/scala/freestyle/module.scala
@@ -98,6 +98,7 @@ object openUnion {
       q"""new {
         ..$parsed
       }"""
+
     c.Expr[Any](expr)
   }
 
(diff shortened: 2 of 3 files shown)


freestyle/shared/src/main/scala/freestyle/debug/package.scala
```diff
@@ -0,0 +1,34 @@
...
+  */
+package object debug {
+  object optionTypes {
+    sealed trait ShowTrees
+  }
+  import optionTypes._
+
+  object options {
+
+    /** Import this value to have Iota print the macro generated code
```
peterneyens: Copy paste error: "Iota" -> "Freestyle"

andyscott: Naturally :)

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -98,6 +102,10 @@ object openUnion {
       q"""new {
         ..$parsed
       }"""
+
+    if (showTrees)
+      c.echo(c.enclosingPosition, s"generated tree:\n${showCode(expr)}")
+
```
peterneyens: As far as I know this `openUnion` macro is only called by the `@module` macro, so that means that we will show this tree twice (which might be confusing).

raulraja (approved):
Cool trick!

freestyle/shared/src/main/scala/freestyle/module.scala
andyscott: Aah-- I'm just getting my grounding so I didn't even look at the macro interplay! I can remove it from `openUnion`.

peterneyens (approved):
Nice :+1:

diesalbla (approved):

andyscott:
I will rework this later this evening. It's a quick adjustment.

anamariamv:
@raulraja @juanpedromoreno @andyscott 
This PR is duplicated in #360.  Could we close it?

Issue #73, opened 2017-02-22
Migrate macros to Scala meta to enable IDEA support


raulraja:
When we started freestyle scala meta and it's macro paradise support had issues with HKTs. Now that scala meta is becoming more stable we should consider dropping the macro code in favor of a version with Scala meta if possible.

https://github.com/scalameta/paradise/issues/59

raulraja:
This has been promoted to very important as it seems IDEA supports scala meta in some capacity and when we move it over it may take care of #58 which is currently blocked.

juanpedromoreno:
Fixed in #344

PR #258, opened 2017-04-28
WIP - Scala Meta

diesalbla:
This PR is a status report on the migration of the macros to `scalameta`, as per issue #73. 

I have translated the `@free` macro to `scalameta`. The present code compiles, and in the annotation code I have indicated how to activate the `scalameta` variant. 

Although the compiles, when it is executed it is raising several errors, due to some aspects of the code not conforming to some invariants of the internal `scalameta` representations.

So far, I have found the following difficulties with `scalameta`: 

- It is not possible to directly use a type name `FF` and write `$FF[_]` inside a quasiquote to indicate a HK parameter, that is why we need a conversion.
- We currently have a problem to unfold a long list of `Decl` or `Defn` inside a quasiquote. In `scala.reflect`, if  `defs : List[DefDef]`, then we could use `trait X {..$defs }`, but now it seems to be more difficult.

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/annotations.scala
+++ b/freestyle/shared/src/main/scala/freestyle/annotations.scala
@@ -17,11 +17,15 @@
 package freestyle
 
 import scala.annotation.{compileTimeOnly, StaticAnnotation}
+// Uncomment [FS-73] lines and comment the rest to switch this macro to scalameta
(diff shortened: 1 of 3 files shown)


diesalbla:
Work on Scalameta has been completed and advanced. Closing this for now.

Issue #254, opened 2017-04-28
Replace `cats.data.Coproduct` and `Inject` with iota to improve interpretation performance.


raulraja:
https://github.com/47deg/iota championed by @andyscott is a faster impl of coproduct based on an indexed List which yields constant access time regardless of the number of Algebras. Let's get this replaced before 0.1.0!

diesalbla:
Replacing `cats` implementation of `Coproduct` and `Inject` (soon to be named `EitherK`) will also allow us to use non-degenerate _unions_ when composing the algebras.

In `cats 0.9.0` (the version we currently use), the `Inject` data type is sealed and restricted to a set of  instances, which follows the instances from the [_Data types à la Carte_](http://dblp.uni-trier.de/rec/bibtex/journals/jfp/Swierstra08) article. These instances restrict the use of the technique to those cases in which the composition of three or more _algebras_, v.g. `A + B + C + D`, is only possible in a right-associative linear way, i.e. `A + (B+ (C + D))`. 

For `freestyle`, this restriction has affected us in the implementation of the `@module` macro. One feature for this annotation is to allow having _modules_ that use other _modules_, e.g.
```scala
@module trait Ann { val x: X ; val y: Y }
@module trait Bob { val z: Z ; val w: W}
@module trait Carl { val a: Ann ; val b: Bob }
```
When expanding the macro for `Carl`, although we can just define `Carl.Op = Coproduct[Ann.Op, Bob.Op]`, this definition is not usable because we then can not build any instance of `Inject[X.Op, Carl.Op]`, as we would need to _lift_ the operations from `X`. 

Instead, our implementation of the `@module` macro needs to assemble the full list `[X,Y,Z,W]` of all algebras, so as to build `Carl.Op = X.Op + (Y.Op + (Z.Op `W.Op) )`. To do this, it needs to perform the recursive search shown [here](https://github.com/47deg/freestyle/blob/7903863839e0d88a09c3a241f59e43c6f49606fd/freestyle/shared/src/main/scala/freestyle/module.scala#L38). This has several drawbacks: 
* It requires us to use type information in the macros. This is more complex than syntactic transformations, and that may be more difficult to do with scalameta #73.  
* It somehow breaks the _encapsulation_ and structuring of code intended for `@module`, since it goes into the inner components of each `@module`. 
[...]

PR #261, opened 2017-04-30
Rework module macro to support targeting iota and cats coproduct types

andyscott:
- Adds Iota 0.0.2 as a dependency
- Introduces `InjK` to handle injecting to Iota and Cats coproducts
- Reworks module macro to target both coproduct types. It would be possible to emit two "Op" types for each algebra, supporting both styles.
- Adjusts docs so that they compile

There's a chance I missed something and the build will fail on CI 😢. 

There's _plenty_ of room for feedback on this. I could use some extra eyes to go through and help find the spots where documentation needs to change.


build.sbt
```diff
@@ -9,6 +9,7 @@ lazy val freestyle = (crossProject in file("freestyle"))
   .settings(name := "freestyle")
   .jsSettings(sharedJsSettings: _*)
   .settings(libraryDependencies ++= Seq(%("scala-reflect", scalaVersion.value)))
+  .settings(libraryDependencies += "com.47deg" %%% "iota-core" % "0.0.2")
```
raulraja: This should go in the 47deg section of sbt-org-policies instead of hardcoded here so it's automanaged by sbt-dependencies and it makes us aware when versions get outdated

docs/src/main/tut/docs/core/modules/README.md
```diff
@@ -105,11 +105,16 @@ for all it's contained algebras. This allows contained algebras to be composed.
 If you were to create this by hand, in the case of the example above, it will look like this:
 
 ```tut:book
-import cats.data.Coproduct
+//import cats.data.Coproduct
 
-type C01[A] = Coproduct[Cache.Op, Database.Op, A]
-type C02[A] = Coproduct[Presenter.Op, C01, A]
-type ManualAppCoproduct[A] = Coproduct[IdValidation.Op, C02, A]
+//type C01[A] = Coproduct[Cache.Op, Database.Op, A]
```
raulraja: Can we remove these commented out snippets?

docs/src/main/tut/docs/core/modules/README.md
```diff
@@ -105,11 +105,16 @@ for all it's contained algebras. This allows contained algebras to be composed.
...
+//type C02[A] = Coproduct[Presenter.Op, C01, A]
+//type ManualAppCoproduct[A] = Coproduct[IdValidation.Op, C02, A]
+
+import iota._
+import KList.:::
+type ManualAppCoproduct[A] = CopK[
```
raulraja: Nice! and much easier to build by hand!

freestyle/shared/src/main/scala/freestyle/InjK.scala
```diff
@@ -0,0 +1,57 @@
...
+
+package freestyle
+
+import cats._
+import cats.arrow.FunctionK
+import cats.free.Inject
+import iota._
+
+/** A generalized injection typeclass to abstract over support for
+  * various coproduct implementations
```
raulraja, line 25: :clap: :tada:

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -142,23 +142,23 @@ object freeImpl {
           import _root_.cats.free.Inject
           import _root_.freestyle.{ FreeS, FSHandler}
 
-          sealed trait $OP[$AA] extends scala.Product with java.io.Serializable
+          sealed trait Op[$AA] extends scala.Product with java.io.Serializable
```
raulraja, line 145: Any reason why we don't use the variable `$OP` instead of just `Op`?

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,188 @@
...
+      classTree       <- makeClassTree(cls)
+      objectTree      <- makeObjectTree(cls, effectTuples)
+    } yield q"$classTree; $objectTree")
+
+  def computeCoproductImpl: Tree =
+    foldAbort(for {
+      tpe             <- compassionateCompanionTypeOf(c.internal.enclosingOwner.owner)
+      fullEffectTypes =  expandEffectType(tpe)
+      fullEffectTrees <- fullEffectTypes.traverse(typeToTypeTree)
+      //coproductTree   <- makeCatsCoproduct(fullEffectTrees) // TODO: make this configurable?
```
raulraja: :fire: https://github.com/47deg/freestyle/issues/262

raulraja (approved):
This is awesome @andyscott , we just need to reflect somewhere in the docs that we use `iota` and the reason perhaps showcase the charts in http://frees.io/docs/core/modules/

freestyle/shared/src/main/scala/freestyle/free.scala
andyscott: I think we should just use the string "Op" because it's a local name; the variable OP can be removed.

docs/src/main/tut/docs/core/modules/README.md
andyscott: Of course!

build.sbt
andyscott: Yep. I wasn't sure exactly what I needed to do and I wanted to get this PR built and passing ASAP.

freestyle/shared/src/main/scala/freestyle/module.scala
andyscott: Should I remove this line? I thought it would be good to keep around for when we feature flag targeting both iota and cats.

raulraja: it's fine to keep, was just pointing out that we are tracking that change

andyscott: Aah 👍 got it.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
-  def impl(c: Context)(annottees: c.Expr[Any]*): c.universe.Tree = {
-    import c.universe._
-    import c.universe.internal.reificationSupport._
+  private[this] def gatherEffects(cls: ClassDef): ValidatedNel[String, List[(Name, Tree)]] =
+    cls.impl.body
+      .collect  { case valDef: ValDef if valDef.rhs.isEmpty => valDef }
+      .traverse { valDef => (valDef.name, valDef.tpt).validNel }
 
-    def fail(msg: String) = c.abort(c.enclosingPosition, msg)
+  private[this] def expandEffectType(tpe: Type): List[Type] = {
```
diesalbla: One point in issue #254 is to avoid deep-collecting algebras through transitive dependencies, which removes the need for this recursive procedure. Instead, we would only need to detect which variables in the annotated trait (or class) are `@free` or `@module`, and then just use only the `Op` types of those variables.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
   }
 
-}
-
-object moduleImpl {
-
-  def impl(c: Context)(annottees: c.Expr[Any]*): c.universe.Tree = {
-    import c.universe._
-    import c.universe.internal.reificationSupport._
+  private[this] def gatherEffects(cls: ClassDef): ValidatedNel[String, List[(Name, Tree)]] =
```
diesalbla: Unless I'm wrong, there seems not to be any case in this procedure that would return an _Invalid_ result. That being so, we could just return the `List`, and just write into the `collect [...] => (valDef.name, valDef.tpt)`.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
-
-      val nbody = body.map {
-        case q"$mods val $name: $eff[..$args]" => q"$mods val $name: $eff[$FF, ..$args]"
-        case x => x
+  private[this] def typeToTypeTree(tpe: Type): ValidatedNel[String, Tree] =
+    Validated.catchNonFatal(c.parse(tpe.toString)).leftMap(t =>
+      NonEmptyList.of(s"unable to convert type $tpe to tree"))
+
+  private[this] lazy val defaultf0: ValidatedNel[String, Tree] =
+    q"""type Op[$AA] = Nothing""".validNel
```
diesalbla: If it fits one line, single quotes is fine. 
Also, ¿why do you use the `Validated`? It seems pretty straightforward.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
 package freestyle
 
+import scala.annotation.tailrec
 import scala.reflect.macros.whitebox.Context
 
-trait FreeModuleLike
-
-object openUnion {
+import cats.data._
+import cats.syntax.all._
```
diesalbla: ¿Can we narrow down the `cats` imports?

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
+    case Nil        => defaultf0
+    case tpt :: Nil => defaultf1(tpt)
+    case _          =>
+      val klist = tpts.foldLeft(tq"$KNil": Tree)((tail, head) =>
+        tq"$KCons[$head.Op, $tail]")
+      q"""type Op[$AA] = $CopK[$klist, $AA]""".validNel
+  }
 
-          type Op[$AA] = $xx.Op[$AA]
+  private[this] def makeClassTree(cls: ClassDef): ValidatedNel[String, Tree] = {
```
diesalbla: ¿Can we narrow down `Tree` to `ClassDef`?

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
+    }
+    ClassDef(mods, name, FFtypeConstructor :: tparams,
+      Template(parents :+ tq"$FreeModuleLike", self, body0)).validNel
+  }
 
-          class To[$LL[_], ..$tts](implicit ..$effArgs) extends $mod[$LL, ..$tns]
+  private[this] def makeObjectTree(
+    cls: ClassDef,
+    effects: List[(Name, Tree)]
+  ): ValidatedNel[String, Tree] = {
```
diesalbla: ¿Can we narrow down `Tree` to `ModuleDef`?

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
 
-          implicit def to[$LL[_], ..$tts](implicit ..$effArgs): To[$LL, ..$tns] = new To[$LL, ..$tns]()
+    val tparams = cls.tparams
```
diesalbla: Here, you can just `import cls.{name, tparams}`.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
-      val effArgs: List[ValDef] = filterEffectVals(userTrait.impl).map( v => toImplArg(v) )
 
-      q"""
-        object ${mod.toTermName} {
+      tail.zipWithIndex.foldLeft(
+        q"""type ${Op(-1)}[$AA] = $Coproduct[$head2.Op, $head1.Op, $AA]""": Tree
+      ) { (prefix, tup) =>
+        val (tpt, i) = tup
+        q"""
+          ..$prefix
```
diesalbla: Since you are not getting any data out from the `prefix`, it seems to me you can just use `map` here.
If you want to return a single `Tree`, you can do the unquoting at the end. However, it may be more informative to return a `List[T]`, where `T` is the actual _type_ in `scala.reflect` of what a member type definition is.

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -16,171 +16,190 @@
...
+      NonEmptyList.of(s"unable to convert type $tpe to tree"))
+
+  private[this] lazy val defaultf0: ValidatedNel[String, Tree] =
+    q"""type Op[$AA] = Nothing""".validNel
+  private[this] def defaultf1(tpt: Tree): ValidatedNel[String, Tree] =
+    q"""type Op[$AA] = $tpt.Op[$AA]""".validNel
+
+  private[this] def makeCatsCoproduct(
+    tpts: List[Tree]
+  ): ValidatedNel[String, Tree] = tpts match {
```
diesalbla: Since there is no _invalid_ case, we can take out the `ValidatedNel`.

diesalbla (commented):
Great work. 
I just left  a few comments and requests for changes on the code.

freestyle/shared/src/main/scala/freestyle/module.scala
andyscott: Sure thing!

freestyle/shared/src/main/scala/freestyle/module.scala
andyscott: Yep-- I have code in Iota I've been prototyping that will help us simplify this method.

freestyle/shared/src/main/scala/freestyle/module.scala
andyscott: Good call. I'm going to leave this code, as it, because it works and _it's not currently used or tested_... so if I changed it I risk breaking it. Once we get feature flagging for targeting cats coproducts, this can be revisited. #262

PR #263, opened 2017-05-01
Fix wording in freestyle-cache documentation

suhasgaddam:
Is the `freestyle-cache` an effect and/or an integration?

It's in the `doc/effects` directory but it's in the `Integrations` navigation menu hierarchy.


raulraja (approved):

raulraja:
It's a type of effect but it has several implementations and also appealed to integrations since it includes a Redis one and that is perhaps more important than the fact that is a type of effect.  I think it can be mentioned in both places, it's generic interface in the effects and it's in memory and Redis impl in the actual integrations

PR #268, opened 2017-05-01
Mre doc edits round 2

MaureenElsberry:
Small language edits s

The Quick Start and main documentation have missing links -- these need to be either explained why they don't exist or added @raulraja from this section:

"Freestyle includes ready to go Algebras and Integrations for the most common application concerns:

Ready to use integrations to achieve parallelism through scala.concurrent.Future, Akka Actors and Monix Task.
Ready to use integrations that cover most of the commons applications concerns such as logging, configuration, dependency injection, persistence, etc.
Traditional effects stacks (reader, writer, state, error, option, either)"


docs/src/main/tut/docs/README.md
```diff
@@ -71,17 +70,17 @@ parametrization of Free programs.
 }
 ```
 
-Freestyle automatically wires all dependencies through implicit evidences that are generated so you don't have to worry about the boilerplate required to build Free based programs.
+Freestyle automatically wires all dependencies through implicit evidences that are generated, so you don't have to worry about the boilerplate required to build Free-based programs.
 
-Once you have these abstract definitions you can combine them in whichever way you want. Freestyle supports nested modules enabling onion style architectures of any arbitrary depth.
+Once you have these abstract definitions, you can combine them in whichever way you want. Freestyle supports nested modules enabling onion-style architectures of any arbitrary depth.
 
-Learn more about [modules](./core/modules) in the extended [documentation]()
```
raulraja, line 78: Is this one of the missing links you were referring to? if so removing it like you did it's fine since we are already linking to the modules section.

raulraja (approved):
Thanks so much @MaureenElsberry, this is great!

PR #273, opened 2017-05-02
Removed ambiguous implicit extensions because FreeS is also a Free

raulraja:


peterneyens (approved):

peterneyens:
It seems you need to change something in the `tagless` tut as well.

PR #295, opened 2017-05-05
Fixed problem with a test application.conf file being included on the freestyle deployed jars…

jrgonzalezg:
… due to a small error on the sbt configuration for freestyle-config. This caused issues with duplicated (and different) application.conf files when trying to package projects that depend on freestyle using tools like sbt-assembly.


raulraja (approved):

raulraja:
Thanks for your contribution @jrgonzalezg !

PR #297, opened 2017-05-07
Integrates sbt-freestyle Plugin

juanpedromoreno:
From now on, Freestyle's SBT configurations will be placed at https://github.com/frees-io/sbt-freestyle , which has been published to Sonatype as snapshot: https://oss.sonatype.org/content/repositories/snapshots/io/frees/


fedefernandez (approved):
Looks great!

PR #298, opened 2017-05-08
Moves integration modules to its Subdirectory

juanpedromoreno:
This PR moves all the integrations modules to its directory, in order to be able to preserve the git history once they are moved to https://github.com/frees-io/freestyle-integrations.

Please, @fedefernandez ^^ . Thanks.


fedefernandez (approved):
giphy 1

PR #310, opened 2017-05-10
Add FunctorK for finally tagless

peterneyens:
This adds [_mainecoon_](https://github.com/kailuowang/mainecoon)'s `FunctorK` and generates instances for `@tagless` traits.

`FunctorK` can be used to execute finally tagless programs in a stack safe way with a `F ~> Free[F, ?]`. I am not sure where to put this implicit, we currently don't really have something in `freestyle-tagless` except the annotation and the macro implementation.


freestyle-tagless/shared/src/main/scala/tagless.scala
```diff
@@ -129,16 +141,32 @@ object taglessImpl {
...
+            FK: _root_.mainecoon.FunctorK[$Eff],
+            fk: _root_.cats.arrow.FunctionK[$MM, $NN]
+          ): $Eff[$NN, ..$TTs] = FK.mapK(h)(fk)
         }
       """
     }
 
-    gen()
+    val output = gen()
+    println(output)
```
raulraja: :fire:

raulraja (approved):
Can the implicits go in thr tagless package object? Ideally we'd put them wherever they are auto imported when the user imports the tagless annotation. We can also have freestyle.tagless.implicits._

peterneyens:
I added a `tagless` package and package object.

freestyle-tagless/jvm/src/test/scala/freestyle/tagless/taglessTests.scala
```diff
@@ -36,7 +36,13 @@ class taglessTestsJVM extends WordSpec with Matchers {
     }
 
     "remain stack safe when interpreted to stack safe monads" in {
-      SOProgram[FreeS[Option, ?]](0).interpret[Option] shouldBe Option(iterations)
+      SOProgram[Free[Option, ?]](0).runTailRec shouldBe Option(iterations)
+
+      type StackSafe[F[_]] = {
+        type λ[α] = Free[F, α]
+      }
```
peterneyens: Is this `StackSafe[F]#λ` something we like? I don't think is easier than `Free[F, ?]`.

If we decide to keep it, I should probably move it to the `tagless` package object.

freestyle-tagless/shared/src/main/scala/freestyle/tagless/package.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package freestyle
+
+import cats.~>
+import cats.free.Free
+
+package object tagless {
+  implicit def freestyleTaglessLiftFree[F[_]]: F ~> Free[F, ?] = λ[F ~> Free[F, ?]](Free.liftF(_))
+
+  type StackSafe[F[_]] = {
```
raulraja, line 25: :heart:

raulraja (approved):

PR #315, opened 2017-05-16
Optimization: Replace FunctionK pattern matching with JVM switch for faster @free programs

raulraja:
Included in this PR
- [x] Optimization for faster FunctionK handlers getting translated to JVM switch statements
- [x] Performance charts that showcase the FunctionK and iota Coproduct optimizations
- [x] A new doc section talking about these optimizations and how the make Freestyle faster than traditional manual encodings frequently used with Free, Inject and Coproduct


andyscott (approved):
Looks good :).

Issue #317, opened 2017-05-18
Allows `FS` type aliases to be used in @module traits


raulraja:
Currently is not possible to use `FS` type aliases inside modules but it should be easy to implement by just having Modules auto extends `EffectLike` in the macro expansion since the carrier `F[_]` it's already present in the expansion.

PR #318, opened 2017-05-18
Allows FS type aliases in @module traits

juanpedromoreno:
Fixes https://github.com/frees-io/freestyle/issues/317

This PR makes that `modules` extends `EffectLike` in the macro expansion, hence, the `FS` type aliases can be used inside them.

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/module.scala
+++ b/freestyle/shared/src/main/scala/freestyle/module.scala
@@ -166,8 +166,9 @@ final class moduleImpl(val c: Context) {
       case q"$mods val $name: $tpt[..$args]" => q"$mods val $name: $tpt[$FF, ..$args]"
       case x => x
     }
+    val parentTrees = parents ++ List(tq"_root_.freestyle.FreeModuleLike", tq"_root_.freestyle.EffectLike[$FF]")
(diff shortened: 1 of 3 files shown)


diesalbla (approved):

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -166,8 +166,9 @@ final class moduleImpl(val c: Context) {
       case q"$mods val $name: $tpt[..$args]" => q"$mods val $name: $tpt[$FF, ..$args]"
       case x => x
     }
+    val parentTrees = parents ++ List(tq"$FreeModuleLike", tq"freestyle.EffectLike[$FF]")
```
raulraja: should these be fully qualified to preserve macro hygiene? `_root_.freestyle.FreeModuleLike` and `_root_.freestyle.EffectLike[$FF]` ?

raulraja (approved):

freestyle/shared/src/main/scala/freestyle/module.scala
juanpedromoreno: Makes sense to me, what do you think @diesalbla ?

diesalbla: We should, yes.

PR #326, opened 2017-05-25
Use nested coproduct list support from Iota 0.2.x

andyscott:
~This build will fail because the docs depend on the integration library which depends on the core macros for generating the free boilerplate. I've added a type to the `@free` generated code, so the integration library build is now outdated.~

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/free.scala
@@ -147,6 +147,7 @@ object freeImpl {
       q"""
         object ${Eff.toTermName} {
 
+          type OpTypes = _root_.iota.KCons[Op, _root_.iota.KNil]
           sealed trait Op[$AA] extends scala.Product with java.io.Serializable {
             val $indexName : _root_.scala.Int
           }
--- a/freestyle/shared/src/main/scala/freestyle/module.scala
+++ b/freestyle/shared/src/main/scala/freestyle/module.scala
@@ -65,23 +65,9 @@ final class moduleImpl(val c: Context) {
       cls             <- decodeAnnotationTarget(annottees.toList)
       effectTuples     = gatherEffects(cls)
       classTree       <- makeClassTree(cls)
-      objectTree      <- makeModuleTree(cls, effectTuples)
+      coproductTree    = makeCoproductTree(effectTuples)
+      objectTree      <- makeModuleTree(cls, coproductTree, effectTuples)
     } yield q"$classTree; $objectTree")
-
-  def computeCoproductImpl: Tree =
-    foldAbort(for {
-      tpe             <- compassionateCompanionTypeOf(c.internal.enclosingOwner.owner)
-      fullEffectTypes =  expandEffectType(tpe)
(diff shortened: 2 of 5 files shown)


freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -100,66 +86,6 @@ final class moduleImpl(val c: Context) {
     cls.impl.body.collect {
       case valDef: ValDef if valDef.rhs.isEmpty => (valDef.name, valDef.tpt) }
 
-  private[this] def expandEffectType(tpe: Type): List[Type] = {
```
raulraja, line 103: 😍

andyscott: total lines changed `+22 −85` 😉

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -171,8 +97,24 @@ final class moduleImpl(val c: Context) {
       Template(parentTrees, self, body0)).validNel
   }
 
+  private[this] def makeCoproductTree(effectTuples: List[(Name, Tree)]): Tree = {
```
diesalbla: * ¿Can we do the `map (_._2)` in the call to `makeCoproductTree`? 
* ¿Can we narrow down the type of the `Tree` to what `.tpt` is?

freestyle/shared/src/main/scala/freestyle/module.scala
```diff
@@ -65,24 +65,10 @@ final class moduleImpl(val c: Context) {
       cls             <- decodeAnnotationTarget(annottees.toList)
       effectTuples     = gatherEffects(cls)
       classTree       <- makeClassTree(cls)
-      objectTree      <- makeModuleTree(cls, effectTuples)
+      coproductTree    = makeCoproductTree(effectTuples)
```
diesalbla: You can move the calls to `gatherEffects` and `makeCoproductTree` inside `makeModuleTree`.

diesalbla (approved):
Great cleanup.

freestyle/shared/src/main/scala/freestyle/free.scala
```diff
@@ -141,6 +141,7 @@ object freeImpl {
 
           import _root_.freestyle.{ FreeS, FSHandler, InjK }
 
+          type OpTypes = _root_.iota.KCons[Op, _root_.iota.KNil]
```
diesalbla: Why do we need to change the `@free` macro? The `Op` is a constant name, not a fresh type name.

andyscott:
Once we eventually have this PR passing, is there an easy way to compile the documentation against this release before we merge (to make sure it's all working 👍 )?

juanpedromoreno:
@andyscott It's already being checked: https://github.com/frees-io/freestyle/blob/feature/iota-nested-coproducts/.travis.yml#L30

juanpedromoreno (approved):

juanpedromoreno:
I'm merging since current issues are related to coursier cache. Checked locally and it's working as expected.

PR #333, opened 2017-05-30
FS-73_1 Migrate `@free` to Scalameta

diesalbla:
This PR contains the first part of issue #73, which is to reimplement the macros using [scalameta](http://scalameta.org/). In this part, we reimplement the `@free` macro in scalameta.

The translation does not allow for a completely 1:1 mapping between the old and the new version. Because Scalameta represents the Scala AST using data classes that do not need the `Context` component, with Scalameta many of the inner classes like `Request` can be moved to the top level (inside the `package`). Furthermore, some utilities that only depend on `Scalameta` datatypes can be moved to a separate file. Since the package `freestyle` has too many members, we have moved the implementation of the `@free` and `@module` macros to an inner `macroimpl` package.

Although we try to use Scalameta quasiquotes as much as possible, these are less flexible than those from `scala.reflect`. For instance, Scalameta does not provide a `cq""` quasiquote for match cases, so we need to either build the `Case` object, or use a quasiquote for a `match`. Another example, taking a type name `F` and writing into a quasiquote `F[_]` is not as flexible unless we can align the actual type of F. Also, we can not unquote the integer variable $indexValue.


build.sbt
```diff
@@ -7,7 +7,7 @@ lazy val root = (project in file("."))
 lazy val freestyle = (crossProject in file("freestyle"))
   .settings(name := "freestyle")
   .jsSettings(sharedJsSettings: _*)
-  .settings(libraryDependencies ++= Seq(%("scala-reflect", scalaVersion.value)))
+  .settings( scalametaSettings)
```
raulraja: whitespace :fire:

build.sbt
```diff
@@ -28,8 +28,9 @@ lazy val tagless = (crossProject in file("freestyle-tagless"))
   .settings(name := "freestyle-tagless")
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(commonDeps: _*)
-  .settings(resolvers += Resolver.bintrayRepo("kailuowang", "maven"))
+  .settings(scalametaSettings)
   .settings(
+    resolvers += Resolver.bintrayRepo("kailuowang", "maven"),
     libraryDependencies += "com.kailuowang" %%% "mainecoon-core" % "0.0.5"
```
raulraja: We should add this sbt-org-policies to track new versions as they are released.

build.sbt
```diff
@@ -28,8 +28,9 @@ lazy val tagless = (crossProject in file("freestyle-tagless"))
   .settings(name := "freestyle-tagless")
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(commonDeps: _*)
-  .settings(resolvers += Resolver.bintrayRepo("kailuowang", "maven"))
+  .settings(scalametaSettings)
   .settings(
+    resolvers += Resolver.bintrayRepo("kailuowang", "maven"),
```
raulraja: Seems like newer versions are available in central and we may not need this additional repo declaration https://mvnrepository.com/artifact/com.kailuowang/mainecoon-core_2.12

freestyle/shared/src/test/scala/freestyle/free.scala
```diff
@@ -77,46 +204,6 @@ class freeTests extends WordSpec with Matchers {
       program.interpret[List] shouldBe List(2)
     }
 
-    "allow multiple args in smart constructors" in {
```
raulraja, line 80: why are all these tests being removed?

diesalbla: They aren't. They are being moved and grouped. This one in particular is now in line 54.

freestyle/shared/src/test/scala/freestyle/free.scala
```diff
@@ -16,26 +16,109 @@
...
+
+    "a trait with some concrete non-FS members" in {
+      """@free trait X {
+        def x: FS[Int]
+        def y: Int = 5
+        val z: Int = 6
+      }""" should compile
+    }
+
+    "a trait with a request with multiple params" in {
```
diesalbla: @raulraja Here it is.

diesalbla:
This PR is to be superseded by another one.

PR #334, opened 2017-05-31
Docs Setup Conditional

diesalbla:
Part of the travis setup, about jekyll and docs, should only be done in the `2.12.2` build.


diesalbla:
@juanpedromoreno This avoids the `install` part that takes half a minute. We could also try to isolate the `docs` part into a separate build. The work specific to the docs-`tut` takes about 5 minutes off the 11-minute build.

juanpedromoreno (approved):

PR #336, opened 2017-06-01
Git Docs: branch and swallow clone

diesalbla:
We change the doc building part: we add a check such that, if
there is a branch in the docs repo with the same name as the current
branch buing built, then we check out that branch.
We also add an option to check out no history from the docs repo.


juanpedromoreno (approved):
👍

PR #337, opened 2017-06-01
FIX: Update iota version to repair builds

diesalbla:
Travis is not building because we are using a iota that does not exist.


build.sbt
```diff
@@ -10,7 +10,7 @@ lazy val freestyle = (crossProject in file("freestyle"))
   .settings(libraryDependencies ++= Seq(%("scala-reflect", scalaVersion.value)))
   .crossDepSettings(
     commonDeps ++ Seq(
-      %("iota-core"),
+      %("iota-core", "0.2.0"), // TODO fold with sbt-org-policies
```
juanpedromoreno, line 13: This is already fixed through sbt-freestyle plugin (and merged in master)

PR #339, opened 2017-06-02
Rearrange interdependencies

diesalbla:
The dependency between tagless and freestyle-core is made only for
tests. We have to add the cats libraries to dependencies.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ lazy val freestyleJVM = freestyle.jvm
 lazy val freestyleJS  = freestyle.js
 
 lazy val tagless = (crossProject in file("freestyle-tagless"))
-  .dependsOn(freestyle)
+  .dependsOn(freestyle % "test -> compile")
   .settings(name := "freestyle-tagless")
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(commonDeps: _*)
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -24,14 +24,16 @@ lazy val freestyleJVM = freestyle.jvm
...
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(commonDeps: _*)
   .settings(resolvers += Resolver.bintrayRepo("kailuowang", "maven"))
-  .settings(
-    libraryDependencies += "com.kailuowang" %%% "mainecoon-core" % "0.0.5"
-  )
+  .settings( libraryDependencies ++= commonDeps ++ Seq(
+    "com.kailuowang" %%% "mainecoon-core" % "0.0.5",
+    %%%("cats-core"),
+    %%%("cats-free")
```
peterneyens: I think these need to be added in `crosDepSettings` ?

Also, if you include `cats-free` you can omit `cats-core`.

peterneyens:
After this PR, we should probably mention in the [tagless docs](https://github.com/frees-io/freestyle-docs/tree/master/src/main/tut/docs/core/tagless) that you need a dependency on freestyle-core as well.

PR #343, opened 2017-06-05
Enables CI checks for Freestyle Projects

juanpedromoreno:
This PR brings a new sh script that clones both projects `docs` and `integrations` within the Travis build space, and makes checks that both pass as it should.


diesalbla (approved):
Good cleanup.

scripts/build.sh
```diff
@@ -0,0 +1,33 @@
...
+
+DOCS_REPO="https://github.com/frees-io/freestyle-docs.git"
+INTEGRATIONS_REPO="https://github.com/frees-io/freestyle-integrations.git"
+
+VERSION="$(grep -F -m 1 'version in ThisBuild :=' version.sbt)"; VERSION="${VERSION#*\"}"; VERSION="${VERSION%\"*}"
+
+echo "Checking projects $DOCS_REPO and $INTEGRATIONS_REPO for freestyle version $VERSION"
+
+set -x
+if [ $(git ls-remote $INTEGRATIONS_REPO | grep "refs/heads/$TRAVIS_BRANCH" | wc -l) = 1 ] ; then
```
diesalbla: If we were to get other repos, I would suggest using a Bash function here.

Issue #314, opened 2017-05-16
can't be used with wartremover


fommil:
sorry, me again, wartremover points out that `@free` generated `case class`es should be final and that implicits in `@module` should have an explicit description (I assume that means type parameter?)

I guess suppressing them is possible, but these ones seem like they could be easily fixed.

raulraja:
Keep them coming, it'd be easier to address all the wart-remover warnings in a single PR since most of them will come from the macros.

diesalbla:
@fommil We have recently released `0.2.0` and updated the macros. To follow up with this issue, ¿could you tell us how we can reproduce the issues you were having?

fommil:
Just add wartremover in your tests. `wartremoverErrors in Test = ...`

fommil:
seeing lots of `StringPlusAny` warnings in 0.2.0, and also `StringContext` need to be in scope to use `@free`

fommil:
and `Throw`.

OK, starting to look like all the warts are brokened :stuck_out_tongue_closed_eyes:

fommil:
hmm, I'm trying by adding this to `tests`

```scala
    wartremoverWarnings := Warts.unsafe ++ Seq(
      Wart.FinalCaseClass,
      Wart.ExplicitImplicitTypes
    ),
```

and this to `project/plugins.sbt`

```
addSbtPlugin("org.wartremover" % "sbt-wartremover" % "2.0.3")
```

but it's not kicking in... hmm

PR #348, opened 2017-06-07
reproduction for #314

fommil:


build.sbt
```diff
@@ -8,6 +8,7 @@ lazy val freestyle = (crossProject in file("freestyle"))
   .settings(name := "freestyle")
   .jsSettings(sharedJsSettings: _*)
   .settings(libraryDependencies ++= Seq(%("scala-reflect", scalaVersion.value)))
+  .settings(wartremoverWarnings in (Test, compile) := Warts.unsafe ++ Seq(Wart.FinalCaseClass, Wart.ExplicitImplicitTypes))
```
fommil, line 11: you could go for `Warts.all` if you want to be super compatible!

PR #349, opened 2017-06-07
Calculates coverage only for JVM, excluding JS tests

juanpedromoreno:
Fixes https://github.com/frees-io/freestyle/issues/347

PR #350, opened 2017-06-08
Calculates coverage only for JVM, excluding JS tests

### LINKED ISSUE #347: Travis Build: Coverage and JS
We have found before that Codecov has some compatibility problems with ScalaJS. As a result, when building the `orgScriptCI` using the ScalaJS backend, we usually find some problems. The goal of this ticket is to avoid this from the build script. 

One possible solution to do this is the following one: 
* Separate `.travis` builds for the JS and the JVM, by using an [environment variable](https://docs.travis-ci.com/user/build-stages/matrix-expansion/) that would split two sub-builds, as we do with versions of Scala.
* In `sbt`, create specialized tasks for `JVM` or `JS`, that only run the usual tasks for the JVM projects or the JS projects. The one for `JS` should not include `codecov`.

juanpedromoreno:
Fixes #347

Also:
* Avoids running ScalaJS tests twice
* Bring matrix expansions to separate ScalaJVM/ScalaJS Travis builds
* `build.sh` now accumulates errors in the exit code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ scala:
 - 2.11.11
 - 2.12.2
 
+env:
+- SCALAENV=jvm
+- SCALAENV=js
+
 jdk:
 - oraclejdk8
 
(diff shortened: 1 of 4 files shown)


juanpedromoreno:
Partially approved in https://github.com/frees-io/freestyle/pull/349

diesalbla (approved):

PR #351, opened 2017-06-08
Refactor Implicits

### LINKED ISSUE #346: Provide trait versions of all implicits.
Often times you want to materialize all implicits under a single import. Currently we are forcing to import each algebra implicit independently which results in many imports in some cases. We should provide both a `trait` and an `object` so users can select their preferred style.

AdrianRaFo:
Fixes #346 Provide trait versions of all implicits.

### DIFF
--- a/freestyle-effects/shared/src/main/scala/effects/validation.scala
+++ b/freestyle-effects/shared/src/main/scala/effects/validation.scala
@@ -18,11 +18,10 @@ package freestyle
 package effects
 
 import cats.{MonadState}
-import cats.data.{State, Validated, ValidatedNel, NonEmptyList}
-
+import cats.data.{NonEmptyList, State, Validated, ValidatedNel}
 
 object validation {
-  final class ValidationProvider[E]{
+  final class ValidationProvider[E] {
     type Errors = List[E]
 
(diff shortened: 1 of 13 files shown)


freestyle-effects/shared/src/main/scala/effects/validation.scala
```diff
@@ -52,27 +50,31 @@ object validation {
         def fromEither[A](x: Either[E, A]): M[Either[E, A]] =
           x match {
             case Left(err) => MS.flatMap(invalid(err))((unit) => MS.pure(x))
```
peterneyens, line 52: Unrelated to this PR, but we can replace `MS.flatMap(...)(_ => MS.pure(x))` with `MS.map(...)(_ => x)` which can also be written as `MS.as(...)(x)`.

Similar in `fromValidatedNel` in the `Invalid` case.

peterneyens (approved):
LGTM. Thanks for cleaning up and improving the code style consistency as well :+1:

Left an unrelated comment.

diesalbla (approved):

raulraja (approved):

raulraja:
Merge pending resolution of https://github.com/frees-io/freestyle/pull/344

PR #352, opened 2017-06-08
Allow optional file and number information in logging calls

### LINKED ISSUE #319: Allow freestyle-logging to capture source file information(class name,line number(?)...) and pass to the logging handler
Possibly using https://github.com/lihaoyi/sourcecode#logging

raulraja:
Fixes #319


freestyle-logging/jvm/src/main/scala/loggingJVM.scala
```diff
@@ -19,35 +19,84 @@ package freestyle
...
 object loggingJVM {
 
   object implicits {
-    implicit def freeStyleLoggingHandler[M[_], C: Manifest](
+    implicit def freeStyleLoggingHandler[M[_]](
         implicit ME: MonadError[M, Throwable]): LoggingM.Handler[M] =
       new LoggingM.Handler[M] {
 
-        val logger = Logger[C]
+        val logger = Logger("")
```
juanpedromoreno, line 31: Rather than leaving this empty string here, I'd provide the class/object name at least.

juanpedromoreno (approved):
LGTM! I left a minor comment/question though

freestyle-logging/jvm/src/main/scala/loggingJVM.scala
raulraja: We can't provide that here because it always infers the handler name and not the class invoking it.
I purposedly added an empty string to avoid showing the same non nonsensical logging we have been seeing until now.

freestyle-logging/js/src/main/scala/loggingJS.scala
```diff
@@ -27,27 +27,73 @@ object loggingJS {
         implicit ME: MonadError[M, Throwable]): LoggingM.Handler[M] =
       new LoggingM.Handler[M] with LazyLogging {
 
-        LoggerConfig.factory = PrintLoggerFactory()
+        private[this] def formatMessage(
+            msg: String,
+            sourceAndLineInfo: Boolean,
+            line: sourcecode.Line,
+            file: sourcecode.File): String =
+          if (sourceAndLineInfo) s"$file:$line: $msg"
```
suhasgaddam, line 35: While this works, it doesn't play well with format strings inside`logback.xml`. https://logback.qos.ch/manual/layouts.html
Saying that, I still haven't figured out how the SLF4J captures the line number...

freestyle-logging/jvm/src/main/scala/loggingJVM.scala
suhasgaddam: It might be possible to pass in the Manifest/ClassTag in each of corresponding methods through the implicit parameter list.

suhasgaddam (approved):

freestyle-logging/shared/src/main/scala/logging.scala
```diff
@@ -21,21 +21,21 @@ object logging {
   @free
   trait LoggingM {
 
-    def debug(msg: String): FS[Unit]
+    def debug(msg: String, sourceAndLineInfo: Boolean = false)(implicit line: sourcecode.Line, file: sourcecode.File): FS[Unit]
```
suhasgaddam, line 24: A comment out of context of this PR.

Because the `msg` parameter is pass by value, there will be a performance hit(string concatenation) if the log level is high/low enough. `verizon/journal` deals with this by having a macro guard the actual `logger.debug` call guarded by adding "if logLevel.isEnabled". Is there a way to do that with the `LoggingM` algebra or to "automatically" lift the `msg` parameter to a `Function0`?

ShaneDelmore: Is it enough just to use call by-name?
Def debug(msg: => String...

suhasgaddam: That's what I originally thought but found something odd.

Switching the `LoggingM` algebra to `def debug(msg: => String): FS[Unit]`
results in a macro expansion of the corresponding ADT(I think this is the right word) to be
```scala
    case class DebugOP extends Op[Unit] with scala.Product with scala.Serializable {
      <caseaccessor> <paramaccessor> val msg: _root_.scala.<byname>[String] = _;
      def <init>(msg: _root_.scala.<byname>[String]) = {
        super.<init>();
        ()
      };
      override val FSAlgebraIndex: _root_.scala.Int = 2
    };
```

And the overall macro expansion creates a compiler error of
```scala
[error] /Users/suhas/github/frees-io/freestyle/freestyle-logging/shared/src/main/scala/logging.scala:24: type mismatch;
[error]  found   : String
[error]  required: => String
[error]     def debug(msg: => String): FS[Unit]
[error]               ^
[error] /Users/suhas/github/frees-io/freestyle/freestyle-logging/shared/src/main/scala/logging.scala:21: type mismatch;
[error]  found   : String
[error]  required: => String
[error]   @free
[error]    ^
```

I think it's related to creating a case class with a by-name parameter. From the  Stack Overflow posts I found, that is not possible.

For some reason, it didn't occur to me, until after reading your comment, that the library's macro expansion could be modified to lift to `Function0` for the case class.

That would allow other `@free` algebras to have call-by-name semantics, which might be useful.

raulraja: @suhasgaddam @ShaneDelmore We can support call by name semantics lifting args to `Function0` and that is probably the way to solve this. @suhasgaddam do you want to create an issue for that? thanks!

raulraja:
Merge pending resolution of https://github.com/frees-io/freestyle/pull/344

raulraja:
@pepegar Can we close this since the commits are going to be included in your PR?

pepegar:
Ok, sure!

PR #363, opened 2017-06-19
Remove MonadError requirement for logging algebra

pvillega:
Restricting logging algebra to require MonadError evidence caused issues when using common Monads like Id in testing, as that monad lacks a MonadError instance.

This fix replaces the MonadError evidence by a simple Monad evidence and expects the effect capturing monad at the end of the stack to handle any exceptions raised during Logging.


diesalbla (approved):

raulraja (approved):

raulraja:
@pvillega thank you so much for your contribution!

PR #366, opened 2017-06-20
reduce deps on scala imports

### LINKED ISSUE #312: FQN needed to Any
this is going to sound really weird, but I am experimenting with flag `-Yno-imports` (stop laughing at the back!) and my (drone-dynamic-agents) tests fail with

```
[error] Handler not found to transform `logic.Deps.Op` to `cats.Id`
```

unless I `import scala.Any`. It would be good if everything was FQN in the macro code.

### LINKED ISSUE #314: can't be used with wartremover

fommil:
- close #312 
- working towards #165
- close #314 (or, at least partially) ... sorry generates a bunch of warnings in your tests where they also break the rules. You can suppress or fix...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,11 @@ lazy val freestyle = (crossProject in file("freestyle"))
   .settings(name := "freestyle")
   .jsSettings(sharedJsSettings: _*)
   .settings(libraryDependencies ++= Seq(%("scala-reflect", scalaVersion.value)))
+  .settings(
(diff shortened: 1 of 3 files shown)


raulraja (approved):
Awesome, thanks @fommil !

Issue #365, opened 2017-06-20
context bounds not allowed in algebras


fommil:
I had this algebra that worked in freestyle 0.2

```scala
  final case class Response[T](header: HttpResponseHeader, body: T)

  @free trait JsonHttpClient {
    def get[B: Decoder](uri: Uri,
                        headers: List[HttpHeader] = Nil): FS[Response[B]]

    // using application/x-www-form-urlencoded
    def postUrlencoded[A: UrlEncoded, B: Decoder](
      uri: Uri,
      payload: A,
      headers: List[HttpHeader] = Nil
    ): FS[Response[B]]
  }
```

the idea being that it can be a simple JSON HTTP client that takes in a circe `Decoder` for responses and responds as domain objects... so the caller can say what they expect and get it.

However, in freestyle 0.3 this fails with a macro error regarding `Decoder` implicit divergence (which I think is a red herring). Converting the context bound notation to use an explicit `implicit` parameter block, I get this warning

```
case classes should have a non-implicit parameter list; adapting to 'case class PostUrlencodedOP()(...)'
```

followed by this error

```
[error] <macro>:27: no arguments allowed for nullary method apply
```

which looks like implicits are just not being handled.

It's very annoying that `scala.meta` doesn't output the location of where the macro is being applied in terms of the file that generated it, because it can be super hard to localise where problems are coming from. @xeno-by can anything be done about that?

By explicit implicits I mean

```scala
  @free trait JsonHttpClient {
    def get[B](
      uri: Uri,
      headers: List[HttpHeader] = Nil
    )(
      implicit decoder: Decoder[B]
    ): FS[Response[B]]

    // using application/x-www-form-urlencoded
    def postUrlencoded[A, B](
      uri: Uri,
      payload: A,
      headers: List[HttpHeader] = Nil
    )(implicit encoder: UrlEncoded[A], decoder: Decoder[B]): FS[Response[B]]
  }
```

suhasgaddam:
https://github.com/frees-io/freestyle/pull/360 to debug the expanded macro in `0.3.1-SNAPSHOT`

~~I have a feeling this is related to flatten on the parameter list.~~

fommil:
I've pushed the code to https://github.com/fommil/drone-dynamic-agents/tree/freestyle-bug just run `sbt compile`

with context bounds

```
{
  trait JsonHttpClient[FF$24[_]] extends _root_.freestyle.internal.EffectLike[FF$24] {
[...]

PR #368, opened 2017-06-20
support implicit parameter list in algebras

fommil:
partial fix for #365 and a failing test for the remainder

### DIFF
--- a/freestyle/shared/src/test/scala/freestyle/free.scala
+++ b/freestyle/shared/src/test/scala/freestyle/free.scala
@@ -129,6 +129,25 @@ class freeTests extends WordSpec with Matchers {
       "implicitly[SCtors1[SCtors1.Op]]" should compile 
     }
 
+    "allow implicit parameters in algebra members" in {
+      """
+        trait Bar[T]
+        @free trait Wibble {
+          def wobble[T](foo: String)(implicit bar: Bar[T]): FS[Unit]
+        }
+      """ should compile
+    }
+
+    // https://github.com/frees-io/freestyle/issues/365
+    "allow context bounds in algebra members" ignore {
+      """
+        trait Bar[T]
+        @free trait Wibble {
+          def wobble[T: Bar](foo: String): FS[Unit]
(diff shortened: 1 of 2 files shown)


freestyle/shared/src/test/scala/freestyle/free.scala
```diff
@@ -129,6 +129,25 @@ class freeTests extends WordSpec with Matchers {
       "implicitly[SCtors1[SCtors1.Op]]" should compile 
     }
 
+    "allow implicit parameters in algebra members" in {
```
diesalbla, line 132: Apart from these test cases, specific for type-classes, we should also add the following test cases:  

* A test case with a simpler implicit parameter: 
```Scala
   "@free trait A { def b(x: Int)(implicit y: Int): FS[Unit]" should compile
```

* A test case with only implicit and no non-implicit parameters: 
```Scala
"@free trait A { def b(implicit y: Int): FS[Unit]" should compile
```

diesalbla (approved):
Nice improvement.  I just left a small comment.
Thank you for your contribution, @fommil.

freestyle/shared/src/test/scala/freestyle/free.scala
fommil: ok, but you've caught me literally as I was about to go to bed so maybe somebody could add these tomorrow? :smile:

fommil:
oh whoops, sorry I didn't update this. @raulraja is this still needed or did your recent PR fix it?

raulraja:
Oh I missed this! I fixed it but did not know you could check for the implicit param the way you did it, so hold on until we fix an issue with CI and I'll add these commits with the extra tests and and your check in addition to what's in #377

raulraja:
Additional tests were added on https://github.com/frees-io/freestyle/blob/08cd997a54ce3c617d4e2c6f48445ec8b43ed0d8/modules/core/shared/src/test/scala/freestyle/free.scala#L66-L95

Issue #369, opened 2017-06-20
macro error: missing parameter type for expanded function


fommil:
the code in this branch/file https://github.com/fommil/drone-dynamic-agents/blob/freestyle-bug-for/src/main/scala/logic.scala

fails with

```
[error] <macro>:17: missing parameter type for expanded function ((x$1: <error>) => x$1: @scala.unchecked)
[error]     }).map((_: @scala.unchecked()) match {
[error]             ^
```

It is expanded into the code at the bottom of this report.

This could well be a `scala.meta` bug because previously the entire `trait` was not being sent to `scala.meta` as input

The shortened version is that this

```scala
  def update(old: WorldView): FS[WorldView] =
    for {
      snap    <- initial
      changed = symdiff(old.alive.keySet, snap.alive.keySet)
    } yield old
```

also fails. My guess is there is a problem with single generator `for` comprehensions.

```scala
{
  trait DynAgents[FF$1[_]] extends _root_.freestyle.internal.EffectLike[FF$1] {
    val d: Drone[FF$1]
    val m: Machines[FF$1]
    def initial: FS[WorldView] = (d.getBacklog |@| d.getAgents |@| m.getManaged |@| m.getAlive |@| m.getTime).map({
      case (db, da, mm, ma, mt) =>
        WorldView(db, da, mm, ma, Map.empty, mt)
    })
    def update(old: WorldView): FS[WorldView] = initial.map(snap => {
      val changed = symdiff(old.alive.keySet, snap.alive.keySet)
      val pending = (old.pending -- changed).filterNot({
        case (_, started) =>
          timediff(started, snap.time) >= 10.minutes
      })
      val update = snap.copy(pending = pending)
      (snap, changed, pending, update)
    }).map((_: @scala.unchecked()) match {
      case (snap, changed, pending, update) => update
    })
    private def symdiff[T](a: Set[T], b: Set[T]): Set[T] = a.union(b) -- a.intersect(b)
    def act(world: WorldView): FS[WorldView] = world match {
      case NeedsAgent(node) =>
        m.start(node).map(_ => {
          val update = world.copy(pending = Map(node -> world.time))
          (_, update)
        }).map((_: @scala.unchecked()) match {
          case (_, update) => update
        })
      case Stale(nodes) =>
        nodes.traverse(m.stop).map(stopped => {
          val updates = stopped.map(_ -> world.time).toList.toMap
          val update = world.copy(pending = world.pending ++ updates)
          (stopped, updates, update)
[...]

PR #370, opened 2017-06-20
failing testcase for comprehension with one generator and an assignment

fommil:
failing testcase for #369

@xeno-by this looks like a `scala.meta` bug... can scala.meta parse something like this?

```scala
          def bar: Option[Double] = for {
            f <- getOptionInt
            bar = f.toDouble
          } yield bar
```

### DIFF
--- a/freestyle/shared/src/test/scala/freestyle/module.scala
+++ b/freestyle/shared/src/test/scala/freestyle/module.scala
@@ -177,6 +177,23 @@ class moduleTests extends WordSpec with Matchers {
         |  val a: P.Y
         |}""".stripMargin should compile
     }
+
+    "allow single generator for comprehensions" in {
+      """
+        @free trait Foo {
+          def foo: FS[Int]
+        }
+        @module trait Bar {
+          val foo: Foo
+
+          def bar: FS[Double] = for {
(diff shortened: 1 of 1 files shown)


xeno-by:
@fommil Scalameta can parse code like this, but the converter (which is part of scalameta/paradise) can't deal with code like this. Unfortunately, this issue is very hard to fix.

The approach that we utilize in scalamacros/scalamacros is much more robust, and we are confident that scalamacros/scalamacros will be able to supersede scalameta/paradise. This is where our macro-related efforts are directed at currently.

We don't have a concrete timeline yet, but I'd expect a scalamacros/scalamacros build that can express the `@free` annotation (along with many other macros) to appear this summer.

/cc @olafurpg

fommil:
thanks @xeno-by ... do you think there is any chance for a workaround in the meantime, e.g. optionally just not parsing method bodies?

xeno-by:
Sadly, I don't think so. One of the peculiarities of the scalameta/paradise approach is eager conversion of ASTs, and that can't be changed without significant changes to meta.

fommil:
damn, I'll just have to use the workaround in the meantime.

PR #377, opened 2017-06-30
Enable implicit args and context bound implicits in @free traits

### LINKED ISSUE #365: context bounds not allowed in algebras

raulraja:
Fixes #365 /cc @fommil

### DIFF
--- a/freestyle/shared/src/main/scala/freestyle/internal/free.scala
+++ b/freestyle/shared/src/main/scala/freestyle/internal/free.scala
@@ -186,12 +186,12 @@ private[internal] case class Algebra(
 
 private[internal] class Request(reqDef: Decl.Def, indexValue: Int) {
 
-  import reqDef.{paramss, tparams}
   import ScalametaUtil._
 
   // Name of the Request ADT Class
   private[this] val reqName: String = reqDef.name.value
   private[this] val req: Type.Name  = Type.Name(reqName.capitalize + "OP")
+  private[this] val cboundPrefix = "__$cbound"
 
   private[this] val res: Type = reqDef.decltpe match {
     case Type.Apply(_, args) => args.last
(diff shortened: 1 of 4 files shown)


andyscott (approved):
LGTM

fommil:
Nice.

freestyle/shared/src/main/scala/freestyle/internal/free.scala
```diff
@@ -201,7 +201,21 @@ private[internal] class Request(reqDef: Decl.Def, indexValue: Int) {
...
+    cbound <- tps.cbounds
+  } yield Term.Param(Nil, Term.fresh(cboundPrefix), Some(Type.Apply(cbound, Seq(Type.Name(tps.name.value)))), None)
+
+  val tparams = reqDef.tparams.map {
+    case p if p.cbounds.nonEmpty => p.copy(cbounds = Nil)
+    case p => p
+  }
+
+  val params: Seq[Term.Param] = reqDef.paramss.flatten.map {
+    case p @ param"..$mods $paramname: $atpeopt = $expropt" /* if p.hasMod(mod"implicit")*/ =>
```
fommil, line 215: I recall Eugene showing me a different way to check for mods. But I can remember, sorry.

raulraja: There is `scalameta-contrib` lib that brings the `hasMod` method but it has also some transitive dependencies we don't want in the core at this point.

Issue #36, opened 2017-01-11
Integrate case classy in config algebras


raulraja:
Case class supports now derivation of codecs for product types so config can be mapped directly to things like classes. On top of that it supports also access to the most common combinators provided by libs like Typesafe Config and a module for shocon wich would allow us to remain compatible with ScalaJS //cc @andyscott

andyscott:
Would you like me to take a stab at this?

raulraja:
@andyscott Yes please! The original intent is to replace this https://github.com/47deg/freestyle/blob/master/freestyle-config/shared/src/main/scala/config.scala with a case classy based implementation that supports an operation for also parsing configs to case classes and still supports the existing algebra operations. Thanks!

andyscott:
I won't be able to get this done by today-- I will work on it over the next few days, though, as I prep some more work for Case Classy!

raulraja:
@andyscott What's the status of this issue? thanks!

anamariamv:
Merged #378

PR #378, opened 2017-07-02
Adding support for case classy

franciscodr:
This pull request adds two new methods in config algebra called `loadAs[T]` and `parseStringAs[T]` that will allow us to decode the content of the config file into a case class of type `T`.

### DIFF
--- a/freestyle-config/src/main/scala/config.scala
+++ b/freestyle-config/src/main/scala/config.scala
@@ -27,6 +27,11 @@ import scala.concurrent.duration._
 
 object config {
 
+  case class ConfigError(msg: String, underlying: Option[Throwable] = None)
+      extends Throwable(msg) {
+    underlying foreach initCause
+  }
+
   sealed abstract class Config {
     def hasPath(path: String): Config.Result[Boolean]
     def config(path: String): Config.Result[Config]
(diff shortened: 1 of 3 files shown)


anamariamv:
This PR fixed #36

freestyle-config/src/main/scala/config.scala
```diff
@@ -44,6 +46,8 @@ object config {
     def load: FS[Config]
     def empty: FS[Config]
     def parseString(s: String): FS[Config]
+    def loadAs[T](implicit dec: ConfigDecoder[T]): FS[Either[DecodeError, T]]
```
raulraja: As of `0.3.1` in #377 this can also be expressed as a context bound `T: ConfigDecoder`

freestyle-config/src/main/scala/config.scala
```diff
@@ -74,6 +78,10 @@ object config {
         def empty: M[Config] = ME.pure(loadConfig(ConfigFactory.empty()))
         def parseString(s: String): M[Config] =
           ME.catchNonFatal(loadConfig(ConfigFactory.parseString(s)))
+        def loadAs[T](decoder: ConfigDecoder[T]): M[Either[DecodeError, T]] =
```
raulraja: We don't need to return `Either[DecodeError, T]]` in these methods since we are in the context of monad error.
We can just return `T` and fold on case classy's result to send the Left to `raiseError` and the Right to `pure`.

raulraja (changes requested):

freestyle-config/src/main/scala/config.scala
raulraja: Or simply use `fromEither` https://github.com/typelevel/cats/blob/8c3898002805a70abfe8bd40d338ecdc5eef450a/core/src/main/scala/cats/ApplicativeError.scala#L123

raulraja (approved):

raulraja:
@franciscodr Thanks! and great work!

Issue #362, opened 2017-06-19
Increase test coverage freestyle config module


dominv:
Increase test coverage freestyle config module

PR #387, opened 2017-07-25
Adding test to check config method for nested configurations

dominv:
This PR solves #362 and increases the freestyle-config module coverage to 94.44% adding a test to check the config method for nested configurations.

@raulraja could you please review?

Thanks!


raulraja (approved):
:+1:

PR #393, opened 2017-08-11
adds commercial support statement

MaureenElsberry:
adds blanket commercial support statement


raulraja:
It should be added also here: https://github.com/frees-io/freestyle-docs/blob/ip-11-implement-html-css/src/main/tut/README.md
In that branch which is the one for the new site revamp

raulraja (approved):

PR #395, opened 2017-08-23
Rename algebra operation suffix from `OP` to `Op`

### LINKED ISSUE #382: Be consistent with the `Op` wording
When developing https://github.com/frees-io/freestyle/pull/353, i needed to pattern match on the cases of the ADT generated by a `@free` algebra. In those cases, the `OP` suffix is added to the name of the method, capitalized.

Also, on `@module` annotated traits, a  `Op` type is used to refer to the operation.

As of my understanding both refer to `operation`, so it would be great to use the same term for both, either `OP` or `Op`. It's not a biggie, but I needed to investigate a bit further asking scalac to expand the macro for me to figure out that in ADT cases, it's `OP`.

peterneyens:
Rename the `OP` algebra operation suffix to `Op` to be consistent with the `Op` type in `@module` (and `@free`) companion objects.

I refactored the loggingJVM handler to use the ideomatic method
freestyle handler instead of pattern matching on the algebra operations.

I suspect we only need to change the handler section of freestyle-docs afterwards.

Resolves #382.

### DIFF
--- a/freestyle-logging/js/src/main/scala/loggingJS.scala
+++ b/freestyle-logging/js/src/main/scala/loggingJS.scala
@@ -25,74 +25,62 @@ object loggingJS {
   trait Implicits {
     implicit def freeStyleLoggingHandler[M[_]](implicit M: Monad[M]): LoggingM.Handler[M] =
       new LoggingM.Handler[M] with LazyLogging {
+        import sourcecode.{File, Line}
 
         private[this] def formatMessage(
             msg: String,
             sourceAndLineInfo: Boolean,
-            line: sourcecode.Line,
-            file: sourcecode.File): String =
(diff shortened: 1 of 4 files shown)


raulraja (approved):

peterneyens:
It is the doc section that I mentioned that is failing on Travis.

raulraja:
We can fix all the docs in one pass before releasing 0.4.0 if you don't want to address it here

pepegar (approved):
Looks much more clear! 👍

peterneyens:
I wouldn't now how to address the docs now. To update the docs I need a freestyle version containing these changes. There is sort of a circular dependency in the Travis builds.

Issue #386, opened 2017-07-20
Rename `freestyle-*` modules to `frees-*`


raulraja:
Action derived from https://github.com/frees-io/freestyle/issues/383 to be applied before 0.4.0

peterneyens:
freestyle, freestyle-integrations and freestyle-docs have been changed. 
Do we leave this issue open until freestyle-kafka, freestyle-cassandra, freestyle-rpc, ... have change as well?

juanpedromoreno:
`freestyle-rpc` has been already renamed 👍

PR #396, opened 2017-08-24
Rename modules to frees-xxx

peterneyens:
I also renamed the freestyle module to frees-core like suggested in the issue.

I moved all the modules to a new modules directory.

I introduced a `module` method similar to the one @andyscott uses in iota and case classy to simplify the creation of cross projects.

The `freestyle-integrations` checks pass and the `freestyle-docs` checks fail on Travis because they use the old `freestyle-xxx` snapshots from sonatype.

First part of #386 (we need to do the same for integrations, cassandra, kafka, rpc, ...).


fedefernandez (approved):
Looks good to me!
The Tut compiling errors will be fixed once we merged this?

peterneyens:
To fix the the tut failures we need to change `freestyle-docs`.

We will also need to change the names of the freestyle projects in sbt-org-policies.

raulraja (approved):

peterneyens:
If we merge this PR and afterwards https://github.com/frees-io/freestyle-integrations/pull/20 and https://github.com/frees-io/freestyle-docs/pull/25 and re-execute the master travis builds afterwards, they should be fine again (fingers crossed).

PR #405, opened 2017-10-03
Remove shapeless as a dependency for the published libraries

andyscott:
Shapeless was only being used for `Lazy` for the legacy coproduct support. We can move the dependency to the test configuration and remove the legacy coproduct support.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ lazy val core = module("core")
     commonDeps ++ Seq(
       %("iota-core"),
       %("cats-free"),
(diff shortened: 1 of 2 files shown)


andyscott:
Locally the presentation compiler test failed for me. It failed on the original master branch as well as this new branch. I may have been running that particular test incorrectly.

peterneyens:
I think it has been failing for a while.
The presentation compiler test isn't enabled in CI. I forgot to create an issue about that.

raulraja (approved):

PR #408, opened 2017-10-08
Update to Cats 1.0.0-MF

peterneyens:
Update Cats to 1.0.0-MF

freestyle-integration and freestyle-docs tests will fail


raulraja (approved):
Outstanding job @peterneyens !. Let's bring `OptionM` back and this is good to go.

PR #422, opened 2017-10-24
Adds a custom deploy job stage for releasing

juanpedromoreno:
This PR uses a new feature (still **beta**) to define job stages in Travis.

https://docs.travis-ci.com/user/build-stages/matrix-expansion/

It would be good to include it in our current pipeline to make faster and better our CI builds, more importantly when we are releasing new `Freestyle` versions.

screenshot 2017-10-24 11 02 08

The `Deploy` stage would depend on the `Test` stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ matrix:
   exclude:
   - scala: 2.11.11
     env: FREESBUILD=docs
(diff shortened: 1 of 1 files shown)


raulraja (approved):
:clap:

PR #425, opened 2017-10-26
Import freestyle slick example

### LINKED ISSUE #421: Move freestyle-slick-example to freestyle
https://github.com/frees-io/freestyle-slick-example can be moved to this repo, together with the `todolist` example.

See: https://github.com/frees-io/freestyle/blob/8575defb206d3ea96f31f09845691f95770e5d3e/build.sbt#L293-L305

as reference.

GMadorell:
Fixes #421

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -277,14 +277,11 @@ lazy val httpClient = module("http-client", subFolder = Some("integrations/http"
   .settings(resolvers += Resolver.jcenterRepo)
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(
-    commonDeps ++ Seq(
-      %("hammock-core", "0.7.0"),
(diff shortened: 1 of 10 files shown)


juanpedromoreno (approved):
Looks great!

Thanks!

peterneyens:
Just a remark for the future, not something to hold back this PR: We might want to move from a postgres DB to using H2, making it easier to use locally. That would also allow us to add some tests (otherwise these examples are going to kill our test coverage).

juanpedromoreno:
Good point @peterneyens , I agree.

Let's address it in a different PR . For now, we could skip coverage from this new example, as we do here for the `todo` list application:

https://github.com/frees-io/freestyle/blob/master/project/ProjectPlugin.scala#L43

PR #427, opened 2017-10-26
Excludes slick example from test coverage

juanpedromoreno:
This PR prevents that slick example affects the code coverage.


fedefernandez (approved):
👍

PR #435, opened 2017-11-08
Adds an implicit conversion for ListenableFuture[Void]

fedefernandez:
Working with the Cassandra Datastax library I've dealt with methods that return a [`CloseFuture`](https://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/CloseFuture.html). This `CloseFuture` is an instance of `ListenableFuture[Void]`.

This PR adds an implicit conversion for transforming a `ListenableFuture[Void]` to `ListenableFuture[Unit]`


juanpedromoreno (approved):
LGTM

PR #439, opened 2017-11-09
Disables tests in deploy stage

juanpedromoreno:
Given the current Travis pipeline, all the tests are being executed before the deploy stage. Hence it doesn't make sense to run again as a part of the release process.

It also upgrades the frees-rpc version.


fedefernandez (approved):
Thanks

PR #442, opened 2017-11-10
Allows higher typed params in algrebra methods

fedefernandez:
Fixes #431 by applying an existential type only when the param type is high bounded.

Thoughts?

PR #444, opened 2017-11-13
Support for upper bounds in algebras type arg declarations

fedefernandez:
This will fix the case where we'll have something like `[T <: SomeType: TypeTag]`, as @muradm requested in [this comment](https://github.com/frees-io/freestyle/pull/442#issuecomment-343752562).

It adds a `type` declaration in the *handler* body like this:

* `type A <: SomeType`

And make the casting in the pattern matching with `A` (instead of `_ <: SomeType`).

The problem with this code is that it's acting only when the param type is high bounded. We should add the `type` declaration for all params. Currently, the only issue is to convert a `Type Param` into a `Type` with scalameta.

### DIFF
--- a/modules/core/shared/src/main/scala/freestyle/internal/free.scala
+++ b/modules/core/shared/src/main/scala/freestyle/internal/free.scala
@@ -109,7 +109,7 @@ private[internal] case class Algebra(
       requests.map {
         _.tparams
           .map(_.tbounds.hi)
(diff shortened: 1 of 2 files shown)


raulraja (commented):
Looks good so far 👏

fedefernandez:
Added some tests, IMO is ready for review. We can improve it in another PR once we know how to convert a `Type.Param` into a `Type.Decl`

raulraja (approved):
:clap:

Issue #374, opened 2017-06-28
Support `FS[_]` params in algebras


raulraja:
We are currently unable to implement methods like `ApplicativeError#recoverWith` because they rely on senf `FS[_]` programs as args and that requires mutually recursive interpreters. We need a fix or workaround for this ASAP as this is currently blocking implementation of other effects and starts affecting users. This is also related to https://github.com/frees-io/freestyle/issues/361
There is `FreeK` which seems to work around this limitation in nutcracker https://github.com/TomasMikula/nutcracker/blob/master/src/main/scala/nutcracker/util/FreeK.scala

anamariamv:
We have paused that for the moment. Peter couldn't find a solution but he will keep thinking about it.

PR #445, opened 2017-11-14
Adds initial Freestyle Cassandra docs.

FPerezP:


docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+> schema, and showing errors at compile-time. 
+ 
+Freestyle-Cassandra provides 2 different APIS, a low level one, letting you define a query and bind 
+values to it, and a String Interpolator one, which allows us to write raw queries and validate them 
+at compile time against a previously defined schema.
+
+In the upcoming sections, we'll take a look at both features and how we can take advantage of them.
+
+## Public APIs
+*Frees-Cassandra* provides a set of algebras to interact with the different pieces of a Cassandra
```
fedefernandez: Maybe its worth to link `algebras` to the freestyle docs (http://frees.io/docs/core/algebras/)

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+
+# References
+
+* [Freestyle](http://frees.io/)
+* [Troy](https://github.com/schemasafe/troy)
+* [Contextual](https://github.com/propensive/contextual)
+* [Datastax Driver](http://docs.datastax.com/en/developer/driver-matrix/doc/common/driverMatrix.html)
+
+[freestyle-cassandra-examples]: https://github.com/frees-io/freestyle-cassandra-example
+[Metrifier]: https://github.com/47deg/metrifier
```
fedefernandez: I think we can remove the `Metrifier` link

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+}
+```
+
+## Freatures
+
+### Low level Queries
+The Low-Level API allows us to define queries with placeholders for query constraints, which can be 
+later bound to the final values, running the query in a safe mode, preventing CQL injection. 
+Frees-Cassandra checks the generated query is a valid one at compile time, throwing errors in case 
+any input value doesn't match with the expected data type.
```
fedefernandez: I'm not sure about it, currently, we don't have type checking at compile time, am I'm missing something?

fedefernandez (changes requested):
Good job, it looks great
I left some minor comments

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+  def read[A](resultSet: ResultSet)(implicit FR: FromReader[A]): FS[A]
+
+  def readOption[A](resultSet: ResultSet)(implicit FR: FromReader[A]): FS[Option[A]]
+
+  def readList[A](resultSet: ResultSet)(implicit FR: FromReader[A]): FS[List[A]]
+
+}
+```
+
+## Freatures
```
juanpedromoreno: Features

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+
+## Public APIs
+*Frees-Cassandra* provides a set of algebras to interact with the different pieces of a Cassandra 
+Service. Each of those algebras represent a public API.   
+
+### ClusterAPI
+It provides methods to open/close connections to a Cassandra instance, load specific keyspaces, or 
+Cassandra configuration.
+
+```
```
juanpedromoreno: This should be a `tut` fragment (using a modifier like`book` or `silent`

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+
+}
+``` 
+
+### SessionAPI
+Provides a way to interact with a proper query. We can both define a query as a template - containing 
+placeholders for real values - or a raw query string. Bear in mind that running a raw query using the 
+SessionAPI directly is unsafe and does not check query correctness at compile time. 
+
+```
```
juanpedromoreno: Same here

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+  def readOption[A](resultSet: ResultSet)(implicit FR: FromReader[A]): FS[Option[A]]
+
+  def readList[A](resultSet: ResultSet)(implicit FR: FromReader[A]): FS[List[A]]
+
+}
+```
+
+## Freatures
+
+### Low level Queries
```
juanpedromoreno: Low level API? 

Then we might have different sections for either DDL or DML

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+The Low-Level API allows us to define queries with placeholders for query constraints, which can be 
+later bound to the final values, running the query in a safe mode, preventing CQL injection. 
+Frees-Cassandra checks the generated query is a valid one at compile time, throwing errors in case 
+any input value doesn't match with the expected data type. 
+
+Let's see how we can use it:
+
+```
```
juanpedromoreno: Same here :)

juanpedromoreno (commented):
Minor comments about compiling every piece of Scala code. Looks good :)

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
-# Cassandra
 
-WIP
\ No newline at end of file
+# Freestyle-Cassandra
+
+[Cassandra] atop **Freestyle** is **`frees-cassandra`**.
+
+Freestyle Cassandra is a purely functional library for interacting with Cassandra supporting both,
+[LowQueryAPI] and [StringQueryInterpolation].
```
raulraja: It will support soon object mapping and other APIs so not sure we need to list those here until they are all complete. I think we should just emphasize that is a Scala Purely Functional driver for Cassandra based on the datastax Java Driver

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+    - [Low level API](#low-level-api)
+    - [String Interpolator](#string-interpolator)
+- [References](#references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## What’s frees-cassandra
+
+[frees-cassandra] is a library to interact with cassandra built atop Free and following the
+[Freestyle] philosophy, being macro-powered. We use Troy under the scenes for modeling purposes
```
raulraja: Troy is an implementation detail and can be mentioned in the credits at the end.

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+Cluster.
+
+## Installation
+
+Add the following resolver and library dependency to your project's build file. 
+
+For Scala `2.11.x` and `2.12.x`:
+
+```scala
+Resolver.bintrayRepo("tabdulradi", "maven")
```
raulraja, line 49: Why is this resolver needed?

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+libraryDependencies += "io.frees" %% "frees-cassandra" % "0.0.4" 
+```
+
+## About Freestyle Cassandra
+
+> [Troy](https://github.com/schemasafe/troy) is a type-safe & compile-time-checked wrapper around
```
raulraja: No need to mention this here as we may actually change this impl down the road and not be based on Troy. We should mention it in the credits or on a section where we describe impl details.

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+To do so, we need to first define an annotated trait specifying a path to the file containing the
+cassandra keyspaces & tables schemas. 
+
+Example
+
+```  @SchemaFileInterpolator("/schema.cql")
+     trait DummySchemaInterpolator
+```
+
+This annotation will be expanded via macros to implement a companion object defining a
```
raulraja: We don't need to specify details as to how the impl of expanding the object works as that may change and this would get outdated. I think here we should just emphasize that you can validate the schema at compile time and runtime and what the user code to do that looks like.

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+```  @SchemaFileInterpolator("/schema.cql")
+     trait DummySchemaInterpolator
+```
+
+This annotation will be expanded via macros to implement a companion object defining a 
+schemaValidator as well as a cql string interpolator method. Now we can import this 
+DummySchemaInterpolator and take advantage of this macro generated utilities [1](#1).  
+
+```
+  import DummySchemaInterpolator._
```
raulraja: If this is expressed inside a @module or @free algebra we don't need to use `FreeS` but instead just `FS` or `FS.Seq`

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+
+  val insertUserTask: FreeS[F, ResultSet] =
+    cql"INSERT INTO demodb.users (id, name) VALUES ($uuid, 'Username');".asResultSet[F]()
+  val getUserTask: FreeS[F, ResultSet] =
+    cql"SELECT id, name FROM demodb.users WHERE id = $uuid".asResultSet[F]()
+```
+
+At the end, what we are getting with `FreeS[F, ResultSet]` is a `Free[FreeApplicative[F, ?], A]`
```
raulraja: The actual value of the alias may change in the future once we figure out how to do recursive free programs as in #374 so we should probably not expand here into what `FreeS` means

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+which can be monadically composed as follows:
+
+```
+for {
+      _             <- insertUserTask
+      userResultSet <- getUserTask
+    } yield userResultSet
+```
+
+Full String Interpolator example is availiable at [Github Frees-Cassandra-Example](https://github.com/frees-io/freestyle-cassandra-example)
```
raulraja: It may be useful to include that under a section in the docs. The different libs like Kafka and Cassandra will eventually have many sections an can potentially grow beyond one menu entry at frees.io since we plan to considerably expand in their functionality and what they can do. Perhaps bringing a couple or one example from there here would be useful for users to understand how to use the interpolator.

docs/src/main/tut/docs/cassandra/README.md
```diff
@@ -4,6 +4,265 @@ title: Cassandra
...
+
+##### 1 Note that we need to define this trait in a different compilation unit (e.g.: a different SBT module) since we are using Contextual Macro Interpolator, so a new macro is defined by our macro.
+
+# References
+
+* [Freestyle](http://frees.io/)
```
raulraja: Here in this section we can add a brief description of how each of these libs are used as part of frees-cassandra

docs/src/main/tut/docs/cassandra/README.md
fedefernandez: This is needed for the Troy schema library

docs/src/main/tut/docs/cassandra/README.md
FPerezP: Yep, hehe, you're right, I got excited too soon 😝

docs/src/main/tut/docs/cassandra/README.md
FPerezP: 👍

FPerezP:
@juanpedromoreno @raulraja I added tut:silent to all code snippets but one, the one containing the `import SchemaInterpolator._`. SchemaInterpolator is decorated with @SchemaFileInterpolator("/schema.cql"), which expands code defining a new macro (using Contextual). It makes the annotated trait to has to be in a different compilation unit, which is not the case using tuts. If you agree, I can create a ticket to investigate how to fix it or implement a workaround. 
There is more info about the macro generated by another macro in this PR: https://github.com/frees-io/freestyle-cassandra-example/pull/6

fedefernandez (approved):

docs/src/main/tut/docs/cassandra/README.md
FPerezP: Thanks @raulraja!I updated the snippets included here in the docs, but also the example I extracted those snippets from. There is a PR for it as well: https://github.com/frees-io/freestyle-cassandra-example/pull/8

docs/src/main/tut/docs/cassandra/README.md
FPerezP: Thanks @juanpedromoreno! At the moment we are only supporting DDL ops. I'm gonna clarify this to avoid confussions. Thanks a lot!

FPerezP: Thanks @juanpedromoreno! At the moment we are supporting both DDL and DML ops, being them transparent to the final user. We are describing here how to interact with datastax objects. I'm not sure if I'm getting you here, sorry. Thanks a lot!

FPerezP:
@juanpedromoreno @raulraja thanks a lot for all your suggestions! They all have been addreesed/commented. Could you please have another look? Thanks!

juanpedromoreno (approved):
Great job @FPerezP !

raulraja (approved):
:clap: let's leave the schema interpolator example without tut, it's not worth going down the rabbit hole to get that working with tut

PR #446, opened 2017-11-14
Support for explicit [F[_]] carrier types in `@module`, `@free` & `@tagless`

### LINKED ISSUE #345: `F[_]` reference in @module, @free and @tagless
There are cases where we need to refer to `F` inside those traits not only at the object level but at the Algebra level as well.
Currently the macro generates freshnames for those but we should keep the `F` convention stable.

raulraja:
Fixes #345 

This PR allows users to add the `F[_]` carrier type and enables better syntax highlighting in IDEA for those that choose to do so.


juanpedromoreno (approved):
Outstanding job @raulraja , LGTM!

PR #447, opened 2017-11-14
Make Algebras and Modules concrete with `.instance`

raulraja:
This PR allows additional syntax in `@free` and `@module` companions that allows obtaining an already concrete instance to the `Op` of the container to be used at the edge of applications. 
Currently uers are forced to use `App[App.Op]` to materialize the `@module trait App` Coproduct. this PR allows them to obtain the same result by simply doing `App.instance` which then they can use to run their effects at the edge of the application.


juanpedromoreno (approved):
:shipit:

andyscott (approved):
👍

PR #451, opened 2017-11-17
Declares all parameter types for the algebra methods

fedefernandez:
This PR adds all type parameters defined in the algebra methods as `type def`, for using them later in the handler matcher.


juanpedromoreno (approved):
LGTM

ghost (approved):

PR #452, opened 2017-11-17
Disable animation when seen once

### LINKED ISSUE #441: Disable main page libraries section animation when seen once
Probably it could be a bit jarring to wait for the animation to complete while you are already navigating on the web.

So we could look for a way to show the animation just once when the user gets into the page, and then just present the content. Given that this is not an SPA, we could look for ways to circumvent the content reloading, maybe a cookie or use the browser's local storage.

calvellido:
* Add session storage data item to control when the animations has been
already played.
* Update anime.js library

This fixes #441


israelperezglez (commented):
LGTM! ;)

PR #458, opened 2017-11-21
Provides async instances to stdlib Future

juanpedromoreno:
This PR provides async instances to go from `scala.concurrent.Future` to any `F`.

### DIFF
--- a/modules/async/async/shared/src/main/scala/async.scala
+++ b/modules/async/async/shared/src/main/scala/async.scala
@@ -16,8 +16,8 @@
 
 package freestyle
 
+import cats.syntax.either._
 import scala.concurrent._
-import scala.util.{Failure, Left, Right, Success}
 
 object async {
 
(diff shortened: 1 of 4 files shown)


modules/async/async/shared/src/main/scala/async.scala
```diff
@@ -33,6 +34,16 @@ object async {
     def async[A](fa: Proc[A]): FS[A]
   }
 
+  class Future2AsyncM[F[_]](implicit AC: AsyncContext[F], E: ExecutionContext)
+    extends FSHandler[Future, F] {
+    override def apply[A](future: Future[A]): F[A] =
+      AC.runAsync { cb =>
+        E.execute(new Runnable {
+          def run(): Unit = future.onComplete(_.toEither)
```
raulraja: Not sure if `Try#toEither` was backported to 2.11. If it works great otherwise pattern matching it is.

raulraja (approved):

modules/async/async/shared/src/main/scala/async.scala
juanpedromoreno: Sadly it is not backported to 2.11.x.

PR #459, opened 2017-11-22
Upgrades to cats 1.0.0-RC1

juanpedromoreno:
It also releases a new patch version.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -279,7 +279,7 @@ lazy val httpClient = module("http-client", subFolder = Some("integrations/http"
   .settings(resolvers += Resolver.jcenterRepo)
   .jsSettings(sharedJsSettings: _*)
   .crossDepSettings(
-    commonDeps ++ Seq(%("hammock-core", "0.7.0"), %("cats-effect") % "test"): _*
+    commonDeps ++ Seq(%("hammock-core", "0.7.1"), %("cats-effect") % "test"): _*
   )
 
 lazy val httpClientJS  = httpClient.js
--- a/build.sbt
+++ b/build.sbt
@@ -211,7 +211,7 @@ lazy val slick = jvmModule("slick", subFolder = Some("integrations"))
 lazy val twitterUtil = jvmModule("twitter-util", subFolder = Some("integrations"))
   .dependsOn(coreJVM)
   .settings(
-    libraryDependencies ++= Seq(%%("catbird-util")) ++ commonDeps
+    libraryDependencies ++= Seq(%%("catbird-util", "0.21.0")) ++ commonDeps
   )
 
 lazy val fetch = module("fetch", subFolder = Some("integrations"))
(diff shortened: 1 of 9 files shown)


fedefernandez:
I think this is ready for review.

I have:
* Upgraded `catbird-util` and `finch-core`
* Updated the travis and sbt files with the latest Scala versions
* Changed the Monix documentation. The instance `Task.nondeterminism` has been removed from Monix. I've added a snippet about how to do parallel processing, but I'm not sure if we should remove the entire section.

@juanpedromoreno @peterneyens @raulraja

modules/core/jvm/src/test/scala/freestyle/nondeterminism.scala
```diff
@@ -49,8 +49,9 @@ class NonDeterminismTests extends FunSuite with Discipline with Matchers {
       }
     }
 
-  implicit val throwableEq: Eq[Throwable] =
-    Eq[String].on(_.toString)
```
peterneyens, line 53: I think this can be `Eq[String].by(_.toString)`.

build.sbt
```diff
@@ -17,7 +17,7 @@ lazy val core = module("core")
   .crossDepSettings(
     commonDeps ++ Seq(
       %("cats-free"),
-      %("iota-core"),
+      %("iota-core", "0.3.3"),
```
raulraja, line 20: Are this behind in org-policies? shouldn't this changes happen there?

modules/tests/src/test/scala/MonixParallelTests.scala
```diff
@@ -18,16 +18,14 @@ package freestyle.tests
...
-import monix.eval.Task.nondeterminism
 import monix.execution.Scheduler.Implicits.global
 import org.scalatest.{Matchers, WordSpec}
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 
 class MonixParallelTests extends WordSpec with Matchers {
   "Applicative Parallel Support" should {
-    "allow non deterministic execution when interpreting to monix.eval.Task" in {
+    "allow non deterministic execution when interpreting to monix.eval.Task" ignore {
```
raulraja, line 28: If this tests are not longer valid we should remove them.

raulraja (approved):

build.sbt
fedefernandez: I've checked it, we can bump the sbt-freestyle plugin to `0.3.11` (we're currently using the `0.3.7` version) but the `0.3.11` uses the `0.3.15` version of `sbt-org-policies` where the selected `iota` version is `0.3.2`.

Two choices:
* Bump `sbt-freestyle` to `0.3.11`
* Release a new version of `sbt-freestyle` and use the new version here

modules/tests/src/test/scala/MonixParallelTests.scala
fedefernandez: The module is not aggregated in the `root` module, should I remove it?

build.sbt
juanpedromoreno: IMHO, let's go for the faster path and let's do the `sbt-org-policies` upgrade after this is out. Makes sense?

fedefernandez:
@peterneyens do you think we can move it forward?

peterneyens:
We have to look a bit closer if we can leverage `Parallel` somehow (like mentioned in #463), but I think we can move forward for now.

fedefernandez:
Sounds good, this PR won't close #463

PR #464, opened 2017-11-27
Frees[F, ?] to M program interpretation implicit def

juanpedromoreno:


peterneyens:
I am not sure if this is an implicit I would put in the default `implicits` object.

In the end this is just an implicit conversion and while that may be convenient in some cases, I am not sure if implicit interpretation is a good idea and adding it to the commonly used `frees.implicits._` import might catch people by surprise. 

We provide some implicit conversions in the `frees-http` extensions but there they are at least a little constrained (you need a `Marshaller`, `EntityEncoder`, ...).

juanpedromoreno:
It makes sense to me. Let me think about where that might fit better, any suggestions are welcome :)

PR #475, opened 2017-12-05
 Moves to freestyle.free

### LINKED ISSUE #469: Move `@free` and `@module` and others under the `freestyle.free` namespace
```
freestyle.@free -> freestyle.free.@free
freestyle.@module -> freestyle.free.@module
package object freestyle -> package freestyle ; package object free
freestyle.implicits -> freestyle.free.implicits
```
The idea of this Pr is to properly namespace `@free` dependent code under the `free` package so we have parity with `freestyle.tagless`.
Additionally all modules and algebras that target only free monads should be namespaced below that as well so we can create counterparts with the same semantics in the `tagless` namespace.

AdrianRaFo:
Fixes #469


raulraja (approved):
:clap:

raulraja:
@juanpedromoreno given the only module failing is the docs and this requires a multi step release, are we ready to merge this and release all other modules?

juanpedromoreno:
Deployment is not going to work, we would need to comment the part of the pipeline where docs are being checked.

juanpedromoreno (approved):

modules/async/async/shared/src/main/scala/free/async.scala
```diff
@@ -0,0 +1,69 @@
+/*
```
juanpedromoreno, line 1: @raulraja @AdrianRaFo Now we have like two `async.scala` files, one inside the free package and another one outside. Is this the expected behavior?

PR #477, opened 2017-12-11
FS-462 Explain FS.Seq and FS.Par type alias.

diesalbla:
Work in progreess. Do not merge. 

This PR resolves 462. We extend the documentation of the `@free` annotation. We explain what `FS`, `FS.Seq` and `FS.Par` are.

We add in the freeetests an example using monadic operations.

### DIFF
--- a/docs/src/main/tut/docs/core/algebras/README.md
+++ b/docs/src/main/tut/docs/core/algebras/README.md
@@ -10,52 +10,38 @@ Algebraic Data Types are the foundation used to define `Free` based applications
 
 When you build an algebra with Freestyle, you only need to concentrate on the API that you want to be exposed as abstract smart constructors, without worrying how they will be implemented.
 
-A trait or abstract class annotated with `@free` or `@tagless` is all you need to create your first algebra with Freestyle:
+In Freestyle, an **algebra** is a trait or abstract class annotated with `@free` or `@tagless`:
 
 ```tut:book
 import freestyle.free._
 
 case class User(id: Long, name: String)
 
-@free trait UserRepository {
+@free trait Users {
   def get(id: Long): FS[User]
   def save(user: User): FS[User]
   def list: FS[List[User]]
 }
 ```
 
- or 
+The `Users` trait declares three smart constructors, named `get`, `save`, and `list`, which generate the basic operations in the algebra. A **smart constructor** is an abstract method declaration with a return type of the form `FS[Ret]`, where `Ret` is the type of the data computed by the operation, and `FS[_]`  marks the method as an operation of the algebra. Intuitively, `FS` means that the method gives a computation within some (generic) context or effect `FS`. 
+For example, the `save` smart constructor has a return type `FS[User]`. Intuitively, this declares `save` as a computation in a context (or effect) `FS`, whose result will be a `User` object. 
 
-```tut:book
-import freestyle.tagless._
+The `@free` is a Scala [macro annotation](https://www.scala-lang.org/blog/2017/10/09/scalamacros.html), that replaces the annotated trait `Users` by a modified trait, and generates a companion object, as in the code below: 
 
-@tagless trait Validation {
-  def minSize(s: String, n: Int): FS[Boolean]
-  def hasNumber(s: String): FS[Boolean]
-}
-
-@tagless trait Interaction {
-  def tell(msg: String): FS[Unit]
-  def ask(prompt: String): FS[String]
-}
-```
-
-
-This is similar to the simplified manual encoding below:
-
-```tut:book
+```tut:silent
 import freestyle.free.FreeS
 import freestyle.free.internal.EffectLike
 
 case class User(id: Long, name: String)
 
-trait UserRepository[F[_]] extends EffectLike[F] {
+trait Users[F[_]] extends EffectLike[F] {
   def get(id: Long): FS[User]
   def save(user: User): FS[User]
   def getAll(filter: String): FS[List[User]]
 }
 
-object UserRepository {
+object Users {
   import _root_.cats.arrow.FunctionK
   import _root_.freestyle.free.InjK
   import _root_.freestyle.free.FreeS
@@ -65,7 +51,7 @@ object UserRepository {
   final case class Save(user: User) extends Op[User]
   final case class GetAll(filter: String) extends Op[List[User]]
 
-  class To[L[_]](implicit ii: InjK[Op, L]) extends UserRepository[L] {
+  class To[L[_]](implicit ii: InjK[Op, L]) extends Users[L] {
 
     private[this] val inj = FreeS.inject[Op, L](ii)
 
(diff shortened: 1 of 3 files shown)


docs/src/main/tut/docs/core/algebras/README.md
```diff
@@ -10,7 +10,7 @@ Algebraic Data Types are the foundation used to define `Free` based applications
 
 When you build an algebra with Freestyle, you only need to concentrate on the API that you want to be exposed as abstract smart constructors, without worrying how they will be implemented.
 
-A trait or abstract class annotated with `@free` or [`@tagless`](../tagless/) is all you need to create your first algebra with Freestyle:
+In Freestyle, you can write an algebra as a trait or abstract class annotated with `@free` (or [`@tagless`](../tagless/)):
```
raulraja: I would not make a (.. distintion here because the next PR with @AdrianRaFo will bring `free` and `tagless` to parity in terms of treatment.

docs/src/main/tut/docs/core/algebras/README.md
```diff
@@ -111,10 +145,41 @@ def myService[F[_]](implicit userRepository: UserRepository[F]) = ???
...
+These type aliases can be used to define some `FS` operations that are derived from other operations. , by combining other operations: 
+
+* `FS.Par[A]` is an alias for `FreeApplicative[F, A]`. You can declare a derived operation of type `FS.Par` by applying the methods of the `Functor` and `Applicative` type classes. 
+* `FS.Seq[A]` is an alias for `Free[FreeApplicative[F, ?], A]`, that can be combined using the methods of the `Monad` type class.
+
+```tut:book
+import cats.syntax.monad._
+@free trait X { 
+  def a: FS[Int]
+  def b(i: Int): FS.Par[Int] = a.map(x => x+i)
```
raulraja: format `x+i` as `x + i` etc to be consistent with the code base and other examples. Sadly we can't run scalafmt in tut code blocks.

raulraja (commented):
:clap: Excellent docs as usual @diesalbla, once we cover the tagless docs in a similar fashion this is ready to go.

diesalbla:
@raulraja For the `@tagless` macro we may need no `Seq` or `Par` type aliases, since we can write  derived operations as follows (an idea by @peterneyens)
```Scala
@tagless trait X {
  def a: FS[Int]
  def b(x: Int): FS[Int]
  def c(implicit f: FlatMap[FS]): FS[Int] = a.flatMap(b)
}
```

On the other hand, the purpose of `Seq` and `Par` was to differentiate combinations using only the `Applicative` (parallel) from those using the full `Monad` (sequential) instance of the target `M`. However, since the point of `@tagless` is that we run each operation into the target `M` directly.

docs/src/main/tut/docs/core/algebras/README.md
```diff
@@ -97,41 +82,102 @@ object UserRepository {
...
+Let us examine how the `@free`-annotated trait relates to the code generated from it. 
+
+### Generalise Trait
+
+The `@free` macro makes two changes to the trait `Users`.
 
-## Automatic method implementations
+1. It adds a type parameter `F[_]`, of kind `* -> *`, which is a general type in which operations are constructed. This makes `Users[F[_]]` resemble a [_Generalised Abstract Data Type (GADT)_](https://pchiusano.github.io/2014-05-20/scala-gadts.html), an interface to construct the algebra's operations into a type `F[_]`.
+
+2. It makes `Users` to extend `EffectLike[F[_]]`. This trait defines the type `FS[_]`, used in the smart constructors, as `FS[_] = FreeS.Par[F]`, which in turn is an alias for `Free[FreeApplicative[F, ?], ?]`.
```
peterneyens: `FreeS.Par[F, A]` is a type alias for `FreeApplicative[F, A]`,
`FreeS[F, A]`(`FS.Seq`) is `Free[FreeApplicative[F, ?], A]`.

Like you mentioned below.

docs/src/main/tut/docs/core/algebras/README.md
```diff
@@ -10,52 +10,38 @@ Algebraic Data Types are the foundation used to define `Free` based applications
...
 ```tut:book
 import freestyle.free._
 
 case class User(id: Long, name: String)
 
-@free trait UserRepository {
+@free trait Users {
   def get(id: Long): FS[User]
   def save(user: User): FS[User]
   def list: FS[List[User]]
```
peterneyens, line 23: Below `list` is `getAll`.

modules/core/shared/src/test/scala/freestyle/free/free.scala
```diff
@@ -80,11 +80,15 @@ class freeTests extends WordSpec with Matchers {
...
+      @free trait X {
         def x: FS[Int]
         def y: Int = 5
         val z: Int = 6
-      }""" should compile
+      }
+      implicit object IdX extends X.Handler[Id]{
+        def x: Int = 31
+      }
+      0 shouldEqual 0
```
peterneyens, line 92: 😄

modules/core/shared/src/test/scala/freestyle/free/free.scala
```diff
@@ -203,12 +214,39 @@ class freeTests extends WordSpec with Matchers {
...
-      object Y extends X.Handler[Id] {
+      implicit object Y extends X.Handler[Id] {
         def a: Int = 5
       }
-      //X[X.Op].b.interpret[Id] shouldEqual 10
+      X.to[X.Op].b.interpret[Id] shouldEqual 10
     }
 
+    "using the Monad instance to combine operations into a FS.Seq" in {
+      import cats.syntax.monad._
```
peterneyens: If it compiles like this, I don't think we need this.
For `flatMap` the import would be `import cats.syntax.flatMap._`.

It might now that it is `Free` which has a `flatMap` itself.

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -178,6 +179,57 @@ class TaglessTests extends WordSpec with Matchers {
...
+      implicit object IdX extends X.Handler[Id] {
+        def a: Int = 41
+      }
+      IdX.b shouldEqual 42
+    }
+
+    "using the Applicative instance to combine operations into a FS.Par" in {
+      @tagless trait X {
+        def a: FS[Int]
+        def b(implicit f: Applicative[FS]): FS[Int] = (a, a).mapN(_+_)
```
peterneyens: `Apply` would do.

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -178,6 +179,57 @@ class TaglessTests extends WordSpec with Matchers {
...
+      }
+      IdX.c shouldEqual 42
+    }
+
+    "mixing all of the above" in {
+      @tagless trait X {
+        def a: FS[Int]
+        def b(i: Int)(implicit f: Functor[FS]): FS[Int] = a.map(x => x+i)
+        def c(implicit ap: Applicative[FS]): FS[Int] = (a,a).mapN(_+_)
+        def d(implicit m: Monad[FS]): FS[Int] = c.flatMap(b)
```
peterneyens: `FlatMap` would be enough.

peterneyens (commented):
Nice job @diesalbla !

juanpedromoreno (approved):
Outstanding job @diesalbla !

raulraja (approved):
Excellent! :clap:

modules/core/shared/src/test/scala/freestyle/free/free.scala
diesalbla: I just want it to compile the code... but maybe it can be better.

PR #479, opened 2017-12-15
Update docs according to object empty in freestyle-rpc

### LINKED ISSUE #457: Update docs according to object empty in freestyle-rpc
According to this PR https://github.com/frees-io/freestyle-rpc/pull/87, we should update the documentation related to RPC.

eperinan:
Added a simple text and example where I try to explain that we can return an empty content as response in a rpc service

It resolves #457


fedefernandez:
Should we add a mention of using `Empty` also as a request?

eperinan:
We support receive an `Empty` as a request. However, IMHO I don't know whether it's a good practice to receive an empty request. 

For the purpose that you mention we could do the next:
`@rpc(Protobuf) def sayHello: FS[Empty.type]``

Is it make sense?

juanpedromoreno:
Could you move this change to Freestyle RPC? Its docs are placed there now. Thanks!

PR #480, opened 2017-12-18
FS-465 Improving config doc

### LINKED ISSUE #465: Config docs improvements
Add docs on loading config to product types as supported by the current case classy integration at http://frees.io/docs/patterns/config/

franciscodr:
Added an example to use config algebra along with case classy library.

It resolves #465

### DIFF
--- a/docs/src/main/tut/docs/patterns/config/README.md
+++ b/docs/src/main/tut/docs/patterns/config/README.md
@@ -39,17 +39,20 @@ sealed trait Config {
   def load: FS[Config]
   def empty: FS[Config]
   def parseString(s: String): FS[Config]
+  def loadAs[T: ConfigDecoder]: FS[T]
+  def parseStringAs[T: ConfigDecoder](s: String): FS[T]
 }
 ```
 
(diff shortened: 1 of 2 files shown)


raulraja:
Looks like we may need to update the CI script?
```
ERROR:  Error installing jekyll:
	The last version of ruby_dep (~> 1.2) to support your Ruby & RubyGems was 1.3.1. Try installing it with `gem install ruby_dep -v 1.3.1` and then running the current command again
	ruby_dep requires Ruby version >= 2.2.5, ~> 2.2. The current ruby version is 2.2.0.
```

franciscodr:
Yes, it seems. I'll update CI script as part of this pull request

docs/src/main/tut/docs/patterns/config/README.md
```diff
@@ -66,6 +69,21 @@ import cats.implicits._
 import scala.util.Try
 ```
 
+We'll create a case class containing all the config values required by our program.
+
+```tut:book
+case class AppConfig(disallowedStates: List[String])
+``` 
+
+Additionally we have to write the config decoder for that case class.
```
raulraja: Do we?, or can this be generically derived automatically?

raulraja (approved):
Would be good to have @MaureenElsberry do a pass since this is all docs related.

MaureenElsberry:
On it!

PR #485, opened 2017-12-22
Update cassandra docs

AdrianRaFo:
Move cassandra docs to frees-cassandra repo and copy them to publish microsite


juanpedromoreno (approved):

project/plugins.sbt
```diff
@@ -1,4 +1,5 @@
 resolvers += Resolver.sonatypeRepo("releases")
 
+addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0")
```
juanpedromoreno: Why are you including coursier?

AdrianRaFo: I need it to download the tut plugin

juanpedromoreno: It's not necessary, `tut` is already being used in several modules.

juanpedromoreno (changes requested):
Let's remove coursier.

juanpedromoreno (approved):

PR #490, opened 2018-01-02
WIP - FS 478 simplify tagless

### LINKED ISSUE #478: Proposal: Simplify Tagless
The `@tagless` macro at present is generating some elements which, in the context of the [tagless-final](http://okmij.org/ftp/tagless-final/) encoding, are redundant.

#### Trait Handler

At present, for the following program: 

```Scala
@tagless trait Foo { 
  def foo(x: Int): FS[Boolean]
}
```

The `@tagless` macro annotation generates the following in the companion object: 
```Scala
trait Foo[F[_]] extends TaglessEffectLike[F] { 
  def foo(x: Int): FS[Boolean]
}

object Foo {
  trait Handler[M[_]] extends Foo[M] {
    def foo(x: Int): M[Boolean]   
  }
  [....]
}
```
I am not sure if this `trait`  definition adds anything, since the `Handler` and the `Foo` contain exactly the same thing. Thus, I would suggest to replace that trait with a type alias
```Scala
object Foo {
   type Handler[M[_]] = Foo[M]
}
```

#### Stack safe Free Handler
One feature of `@taglesss` is to use `@free` annotation to generate a stack-safe interpretation, into the `Free` monad itself. Now `@free` generates a companion object for the Stacksafe algebra, and this later companion also contains a `Handler` definition. And, finally, `@tagless` itself provides a default implementation of this handler. 

To make this clear, expanding both `@tagless` and the inner `@free` would leave this code (abridged): 
```Scala
trait Foo[F[_]] extends TaglessEffectLike[F] { 
  def foo(x: Int): FS[Boolean]
}

object Foo {
  trait Handler[M[_]] extends Foo[M] {
    def foo(x: Int): M[Boolean]   
  }

  trait StackSafe[F[_]] extends FreeEffectLike[F] {
    def foo(x: Int): FS[Boolean]
  }
  object StackSafe {
    [....]
    trait Handler[M[_]] extends (Op ~> M) {
      def foo(x: Int): M[Boolean]     
    }
  }

  implicit def stackSafeHandler[M[_]](implicit ev: Monad[M], hh: Handler[M]): StackSafe.Handler[M] = 
    new StackSafe.Handler[M] {
      def foo(x: Int): M[Boolean]  = hh.foo(x)
    }
  [....]
}
```
I think that the same effect can be achieved, with a bit less hastle, but just making the extended trait 
The same effect would be achieved if we could add an inheritance to the initial `Foo` trait, like follows: 

```Scala
trait Foo[F[_]] extends TaglessEffectLike[F] with Foo.StackSafe.Handler[F] { 
  def foo(x: Int): FS[Boolean]
}
```
Together with the previous alias, we could now remove the `stackSafeHandler` method itself (the implicit `Handler` would become `implicit Foo`, which itself would be `StackSafe.Handler`·

diesalbla:
DO NOT MERGE - WORK IN PROGRESS

Resolves #478

### DIFF
--- a/modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
+++ b/modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
@@ -34,6 +34,107 @@ import modules._
 
 class TaglessTests extends WordSpec with Matchers {
 
+  "the @tagless macro annotation should be accepted if it is applied to" when {
+
+    "a trait with at least one request" in {
+      "@tagless trait X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an F[_] bound type param" in {
+      "@tagless @debug trait FBound[F[_]] { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "an abstract class with at least one request" in {
+      "@tagless abstract class X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an abstact method of type FS" in {
+      "@tagless trait X { def f(a: Char) : FS[Int] }" should compile
+    }
+
+    "a trait with type parameters" ignore {
+      "@tagless trait X[A] { def ix(a: A) : FS[A] }" should compile
+    }
+
+    "a trait with some concrete non-FS members" ignore {
+      """@tagless trait X {
+        def x: FS[Int]
+        def y: Int = 5
+        val z: Int = 6
+      }""" should compile
+    }
+
+    "a trait with a request with multiple params" in {
+      "@tagless trait X { def f(a: Int, b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with a currified request, with multiple params lists" ignore {
+      "@tagless trait X { def f(a: Int)(b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with some implicit parameters, with many params lists" ignore {
+      "@tagless trait X { def f(a: Int)(implicit b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with a request with no args" ignore {
+      "@tagless @debug trait X { def f: FS[Int] }" should compile
+    }
+
+    "a trait with type parameters in the method" in {
+      "@tagless trait X { def ix[A](a: A) : FS[A] }" should compile
+    }
+
+    "a trait with high bounded type parameters in the method" in {
+      "@tagless trait X { def ix[A <: Int](a: A) : FS[A] }" should compile
+    }
+
+    "a trait with lower bounded type parameters in the method" in {
+      "@tagless trait X { def ix[A >: Int](a: A) : FS[A] }" should compile
+    }
+
+    "a trait with different type parameters in the method" in {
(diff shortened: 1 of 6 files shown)


modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -34,6 +34,107 @@ import modules._
...
 class TaglessTests extends WordSpec with Matchers {
 
+  "the @tagless macro annotation should be accepted if it is applied to" when {
+
+    "a trait with at least one request" in {
+      "@tagless trait X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an F[_] bound type param" in {
+      "@tagless @debug trait FBound[F[_]] { def bar(x:Int): FS[Int] }" should compile
```
juanpedromoreno:
```diff
- @tagless @debug trait FBound[F[_]] { def bar(x:Int): FS[Int] }
+ @tagless @debug trait FBound[F[_]] { def bar(x:Int): F[Int] }
```

Shouldn't be in that way?

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -34,6 +34,107 @@ import modules._
...
+
+    "an abstract class with at least one request" in {
+      "@tagless abstract class X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an abstact method of type FS" in {
+      "@tagless trait X { def f(a: Char) : FS[Int] }" should compile
+    }
+
+    "a trait with type parameters" ignore {
```
juanpedromoreno, line 55: Are we ready to don't ignore these tests?

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -34,6 +34,107 @@ import modules._
...
+      "@tagless trait X[A] { def ix(a: A) : FS[A] }" should compile
+    }
+
+    "a trait with some concrete non-FS members" ignore {
```
juanpedromoreno, line 59: For F-bounded, could be this possible?

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
diesalbla: No. Request methods in a `@free` or `@tagless` are identified by the use of the  `FS` type constructor, to distinguish them from the non-request methods.

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
diesalbla: No. There is still some problem with the `@free` macro, that makes it not possible to add type parameters in the trait. This first was detected when we wanted to remove the path-dependent types used in the effects.

diesalbla:
Closing this PR to reopen it with more description.

PR #491, opened 2018-01-03
FS-478 Simplify Tagless

### LINKED ISSUE #478: Proposal: Simplify Tagless

diesalbla:
Resolves #478. Succeeds #490.
 
We simplify the generation of the `Handler`, and of the `StackSafe.Handler` trait definitions. Instead, we make the `Handler` to be a sub-trait of both of them. This removes some code generation

We also change how we generate and use `FunctorK` type-class from the [mainecoon](https://github.com/kailuowang/mainecoon) library. 
* We add to every `@tagless`-annotated trait a `mapK` function, which takes a `FunctionK`. 
* We change the `FunctorK`instance to use the trait's `mapK` function. 
* We change the `implicit def derive` method to use the trait's `mapK` function. 

We move some logic for manipulating Scalameta case classes into the `ScalametaUtil` object. 

We extend the test suite on the `@tagless` macro, copying some tests for the `@free` macro.

### DIFF
--- a/modules/core/shared/src/main/scala/freestyle/free/internal/free.scala
+++ b/modules/core/shared/src/main/scala/freestyle/free/internal/free.scala
@@ -66,7 +66,7 @@ object freeImpl {
 
 }
 
-private[internal] case class Algebra(
+case class Algebra(
     mods: Seq[Mod],
     name: Type.Name,
     tparams: Seq[Type.Param],
--- a/modules/core/shared/src/main/scala/freestyle/free/internal/free.scala
+++ b/modules/core/shared/src/main/scala/freestyle/free/internal/free.scala
@@ -66,7 +66,7 @@ object freeImpl {
 
 }
 
-case class Algebra(
+private[freestyle] case class Algebra(
     mods: Seq[Mod],
     name: Type.Name,
     tparams: Seq[Type.Param],
--- a/modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
+++ b/modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
@@ -34,6 +34,107 @@ import modules._
 
 class TaglessTests extends WordSpec with Matchers {
 
+  "the @tagless macro annotation should be accepted if it is applied to" when {
+
+    "a trait with at least one request" in {
+      "@tagless trait X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an F[_] bound type param" in {
+      "@tagless @debug trait FBound[F[_]] { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "an abstract class with at least one request" in {
+      "@tagless abstract class X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an abstact method of type FS" in {
+      "@tagless trait X { def f(a: Char) : FS[Int] }" should compile
+    }
+
+    "a trait with type parameters" ignore {
+      "@tagless trait X[A] { def ix(a: A) : FS[A] }" should compile
+    }
+
+    "a trait with some concrete non-FS members" ignore {
+      """@tagless trait X {
+        def x: FS[Int]
+        def y: Int = 5
+        val z: Int = 6
+      }""" should compile
+    }
+
+    "a trait with a request with multiple params" in {
+      "@tagless trait X { def f(a: Int, b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with a currified request, with multiple params lists" ignore {
+      "@tagless trait X { def f(a: Int)(b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with some implicit parameters, with many params lists" ignore {
+      "@tagless trait X { def f(a: Int)(implicit b: Int): FS[Int] }" should compile
+    }
+
+    "a trait with a request with no args" ignore {
+      "@tagless @debug trait X { def f: FS[Int] }" should compile
+    }
+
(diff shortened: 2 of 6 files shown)


modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -34,6 +34,107 @@ import modules._
...
 class TaglessTests extends WordSpec with Matchers {
 
+  "the @tagless macro annotation should be accepted if it is applied to" when {
+
+    "a trait with at least one request" in {
+      "@tagless trait X { def bar(x:Int): FS[Int] }" should compile
+    }
+
+    "a trait with an F-Bound type param" in {
+      "@tagless trait FBound[F[_]] { def bar(x:Int): FS[Int] }" should compile
```
raulraja: We should probably also support `F[Int]` since in this case it's the same as `FS[Int]`
```scala
@tagless trait FBound[F[_]] { def bar(x:Int): F[Int] }
```

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
```diff
@@ -34,6 +34,107 @@ import modules._
...
+    }
+
+    "a trait with different type parameters in the method" in {
+      "@tagless trait X { def ix[A <: Int, B, C >: Int](a: A, b: B, c: C) : FS[A] }" should compile
+    }
+
+    "a trait with high bounded type parameters and implicits in the method" in {
+      """
+        trait X[A]
+        @tagless trait Y { def ix[A <: Int : X](a: A) : FS[A] }
```
raulraja, line 107: Do we also support implicit arguments beside context bounds?

raulraja (approved):
We should consider perhaps in a different PR if we can somehow embed the optimization generalization [here](https://typelevel.org/blog/2017/12/27/optimizing-final-tagless.html) in an easy way for users to be able to optimize programs written with `@tagless` in Freestyle

modules/core/shared/src/main/scala/freestyle/free/internal/free.scala
```diff
@@ -66,7 +66,7 @@ object freeImpl {
 
 }
 
-private[internal] case class Algebra(
```
AdrianRaFo, line 69: Maybe this can be kept as `private[freestyle]`

AdrianRaFo (approved):

modules/core/shared/src/test/scala/freestyle/tagless/TaglessTests.scala
diesalbla: There is a test with `implicit` arguments, yes: https://github.com/frees-io/freestyle/pull/491/files/91e17d1a7c0b09def6e51ff1be2a7dec3131c955#diff-f0ef9c02cc3b2899bf09ce3ea089ce66R75

raulraja (approved):
excellent work @diesalbla ! @AdrianRaFo you may want to include this changes in your PR for docs to ensure the docs compile with these changes.

PR #496, opened 2018-01-09
FS-495 Avoid flattening lists of parameters

### LINKED ISSUE #495: Avoid Flattening the list of parameters
The `@free` macro has to transform any request method into a case class in the ADT, a smart constructor in the To class, and a method in the `Handler` trait. A notable change done by this transformation is to _flatten_  the list of parameters in the request. 

However, this transformation makes it difficult to deal with implicit parameters in a method, including those in a type-parameter context bound. 

The goal of this issue is to modify the `@free` (and `@tagless`) macro so that they preserve the lists of parameters, and to add more unit tests to check several cases for this transformation.

diesalbla:
Resolves #495.

* We modify the `@free` macro, to avoid mixing or flattening separate lists of parameters. 

* We increase test coverage for `@free` programs these programs. We avoid using `should compile` in these tests.

### DIFF
--- a/modules/core/shared/src/test/scala/freestyle/free/free.scala
+++ b/modules/core/shared/src/test/scala/freestyle/free/free.scala
@@ -18,6 +18,7 @@ package freestyle
 package free
 
 import cats.{Id, Monoid}
+import cats.kernel.instances.int._
 import cats.instances.option._
 import cats.instances.list._
 import cats.syntax.apply._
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/test/scala/freestyle/free/free.scala
```diff
@@ -132,13 +188,15 @@ class freeTests extends WordSpec with Matchers {
...
-      """ should compile
+      import cats.syntax.apply._
+      @free trait X {
+        def a: FS[Int]
+        def b: FS.Par[Int] = (a, a).mapN(_+_)
+      }
+      object Y extends X.Handler[Id] {
+        def a: Int = 5
+      }
+      //X[X.Op].b.interpret[Id] shouldEqual 10
```
juanpedromoreno, line 209: Is this comment useful?

juanpedromoreno (approved):
Looks good.

modules/core/shared/src/test/scala/freestyle/free/free.scala
diesalbla: No, it's a typo. It should be a line to run the derived `b` method with the `Y` Handler.

juanpedromoreno (approved):

PR #503, opened 2018-01-11
free.LoggingM as tagless.LoggingM.StackSafe alias

peterneyens:
- `freestyle.free.LoggingM` becomes a type alias for `freestyle.tagless.LoggingM.StackSafe`
- Moved `js`/`jvm` `logging` object to `shared`


fedefernandez (approved):
Looks neat! Thanks @peterneyens

PR #507, opened 2018-01-12
Freestyle Effects for @tagless

juanpedromoreno:
This PR brings support for freestyle-effects for @tagless final encoding.

From now on, `asyncCatsEffect` and `asyncGuava` have been moved to `freestyle.async` (both async integrations are not really tied to `free`.

Releases a new minor version due to the breaking changes in this PR: **0.6.0**.

### DIFF
--- a/modules/async/async/shared/src/main/scala/async.scala
+++ b/modules/async/async/shared/src/main/scala/async.scala
@@ -60,7 +60,10 @@ object async {
 
     final class FutureOps[A](f: Future[A]) {
 
+      @deprecated("Use unsafeTo instead.", "0.5.4")
       def to[F[_]](implicit AC: AsyncContext[F], E: ExecutionContext): F[A] = future2AsyncM[F, A](f)
+
+      def unsafeTo[F[_]](implicit AC: AsyncContext[F], E: ExecutionContext): F[A] = future2AsyncM[F, A](f)
 
     }
 
--- a/modules/async/async/shared/src/main/scala/async.scala
+++ b/modules/async/async/shared/src/main/scala/async.scala
@@ -29,6 +29,17 @@ object async {
     def runAsync[A](fa: Proc[A]): M[A]
   }
 
+  def future2AsyncM[F[_], A](
+      future: Future[A])(implicit AC: AsyncContext[F], E: ExecutionContext): F[A] =
+    AC.runAsync { cb =>
+      E.execute(new Runnable {
+        def run(): Unit = future.onComplete {
+          case Failure(e) => cb(Left(e))
+          case Success(r) => cb(Right(r))
+        }
+      })
+    }
+
   trait Implicits {
     implicit def futureAsyncContext(implicit ec: ExecutionContext) = new AsyncContext[Future] {
       def runAsync[A](fa: Proc[A]): Future[A] = {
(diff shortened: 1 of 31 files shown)


raulraja (approved):

modules/effects/shared/src/main/scala/tagless/effects/either.scala
```diff
@@ -0,0 +1,55 @@
...
+    @tagless sealed trait EitherM {
+      def either[A](fa: Either[E, A]): FS[A]
+      def error[A](e: E): FS[A]
+      def catchNonFatal[A](a: Eval[A], f: Throwable => E): FS[A]
+    }
+
+    trait Implicits {
+      implicit def freeStyleEitherMHandler[M[_]](
+          implicit ME: MonadError[M, E]): EitherM.Handler[M] = new EitherM.Handler[M] {
+        def either[A](fa: Either[E, A]): M[A] = fa.fold(ME.raiseError[A], ME.pure[A])
```
peterneyens: This can be `ME.fromEither(fa)`

modules/effects/shared/src/main/scala/tagless/effects/error.scala
```diff
@@ -0,0 +1,46 @@
...
+    def either[A](fa: Either[Throwable, A]): FS[A]
+    def error[A](e: Throwable): FS[A]
+    def catchNonFatal[A](a: Eval[A]): FS[A]
+  }
+
+  trait Implicits {
+
+    implicit def freeStyleErrorMHandler[M[_]](
+        implicit ME: MonadError[M, Throwable]): ErrorM.Handler[M] = new ErrorM.Handler[M] {
+      def either[A](fa: Either[Throwable, A]): M[A] = fa.fold(ME.raiseError[A], ME.pure[A])
```
peterneyens: This could also be `ME.fromEither(fa)`.

fedefernandez (approved):

modules/async/async/shared/src/main/scala/async.scala
```diff
@@ -42,5 +53,18 @@ object async {
...
     }
   }
+
+  trait Syntax {
+
+    implicit def futureOps[A](f: Future[A]): FutureOps[A] = new FutureOps(f)
+
+    final class FutureOps[A](f: Future[A]) {
+
+      def to[F[_]](implicit AC: AsyncContext[F], E: ExecutionContext): F[A] = future2AsyncM[F, A](f)
```
peterneyens: Personally I think I would call this `unsafeTo`, but the `AsyncM` type class has no real "suspension" guarantees.
We could implement it more safely for `Eval[Future[A]]` or a method which takes `=> Future[A]`.

We just have to remember that when we want to end up in something like monix `Task` or `cats.effect.IO`, this `to` method is probably not the best route.

juanpedromoreno: Sounds good, I've just deprecated it in https://github.com/frees-io/freestyle/pull/507/commits/9acae08d059f3cea1364018bfec7ef3e53936fd2

fedefernandez: Sounds good. Created #509

peterneyens (approved):

PR #488, opened 2017-12-28
Equality for tagless and update Quick Start

AdrianRaFo:
Fixes #476 
Fixes #486

PR #514, opened 2018-01-23
Remove Tagless doc. link from menu

calvellido:
* Remove the link to Tagless from the docs sidebar menu, as the information related to it, it is now integrated in other sections, per this PR: https://github.com/frees-io/freestyle/pull/488


eperinan (approved):
Apart from that, IMHO, we should create a redirect from /docs/core/tagless to the new path in order to redirect the users whose find this old path in google.

What do you think?

calvellido:
That would be great, but I doubt we can do it as I think that the info is now spread apart between different sections in the documentation, I could be wrong though.

PR #522, opened 2018-01-29
FS-517 Isues with StackSafe Expansion

### LINKED ISSUE #517: @tagless issue with StackSafe expansion
The following code doesn't compile:

```scala
object test {
  @tagless trait TestF[F[_]] {
    def foo(f: Int => F[String]): F[String]
  }
}
```

It produces:

```bash
[error] <macro>:13:42: not found: type F
[error]       final case class AOp(val f: Int => F[String]) extends _root_.scala.AnyRef with Op[String] { override val FSAlgebraIndex6: _root_.scala.Int = 0 }
[error]                                          ^
[error] <macro>:13:42: not found: type F
[error]       final case class AOp(val f: Int => F[String]) extends _root_.scala.AnyRef with Op[String] { override val FSAlgebraIndex6: _root_.scala.Int = 0 }
[error]                                          ^
[error] <macro>:16:41: not found: type F
[error]         protected[this] def a(f: Int => F[String]): MM$12[String]
[error]                                         ^
[error] <macro>:27:34: not found: type F
[error]         override def a(f: Int => F[String]): FS[String] = toInj10(AOp(f))
[error]                                  ^
[error] <macro>:13:42: not found: type F
[error]       final case class AOp(val f: Int => F[String]) extends _root_.scala.AnyRef with Op[String] { override val FSAlgebraIndex6: _root_.scala.Int = 0 }
```

diesalbla:
Resolves #517. 

There are some issues in `@tagless` when using stacksafe, which are due to the fact that the tagless is more flexible than the `@free` macro.

To avoid these issues, we have separated a `@stacksafe` annotation for the `@tagless` annotated traits. Without it, it does not generate the Stacksafe trait and object.

We keep the annotation in all places where `@tagless` was used before. 

Pending issues:
* [ ] Tests to check that the non-stacksafe macro is more flexible, by using `FS` inside method parameters or return type. 
* [ ] Update documentation to indicate `@stacksafe` as an optional feature. 

Note that this change breaks compatibility: any trait annotated with `@tagless` using the previous algebra will lose the `StackSafe` trait-object pair in the companion object.

### DIFF
--- a/docs/src/main/tut/docs/core/algebras/README.md
+++ b/docs/src/main/tut/docs/core/algebras/README.md
@@ -201,12 +201,12 @@ Freestyle allows us to compose `@free` and `@tagless` algebras. Let us consider 
 ```tut:book
 import freestyle.tagless._
 
-@tagless trait Validation {
+@tagless @stacksafe trait Validation {
   def minSize(s: String, n: Int): FS[Boolean]
   def hasNumber(s: String): FS[Boolean]
 }
 
-@tagless trait Interaction {
+@tagless @stacksafe trait Interaction {
   def tell(msg: String): FS[Unit]
   def ask(prompt: String): FS[String]
 }
--- a/docs/src/main/tut/docs/core/handlers/README.md
+++ b/docs/src/main/tut/docs/core/handlers/README.md
@@ -154,12 +154,12 @@ Tagless final algebras are declared using the `@tagless` macro annotation.
 
 ```tut:book
 
-@tagless trait Validation {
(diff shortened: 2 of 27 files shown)


raulraja (approved):
Excellent @diesalbla :clap:

juanpedromoreno:
Outstanding job @diesalbla 

Could you add specific tests to prove the following code work?

```scala
object test {
  @tagless trait TestF[F[_]] {
    def foo(f: Int => F[String]): F[String]
  }
}
```

Thanks!

PR #523, opened 2018-01-29
Fix syntax highlighting

ioleo:
Due to extra newline, the `tut:book` syntax highlighting directive is shown as part of code example:

screenshot


juanpedromoreno (approved):
@loostro thanks for your time!

PR #529, opened 2018-02-01
FS-528 Override MapK for tagless-stacksafe Handler.

diesalbla:
Override the `mapK` method for the Handler of a tagless-stacksafe algebra.

@loostro

### DIFF
--- a/modules/core/shared/src/test/scala/freestyle/tagless/StacksafeTests.scala
+++ b/modules/core/shared/src/test/scala/freestyle/tagless/StacksafeTests.scala
@@ -263,28 +263,23 @@ class TaglessStacksafeTests extends WordSpec with Matchers {
 
   }
 
-  "Tagless final algebras" should {
-
-    "Allow a trait with an F[_] bound type param" in {
-      "@tagless @stacksafe trait X[F[_]] { def bar(x:Int): FS[Int] }" should compile
-    }
-
-    "combine with other tagless algebras" in {
-
-      def program[F[_] : Monad : TG1 : TG2: TG3] = {
-        val a1 = TG1[F]
-        val a2 = TG2[F]
-        val a3 = TG3[F]
-        for {
-          a <- a1.x(1)
-          b <- a1.y(1)
-          c <- a2.x2(1)
-          d <- a2.y2(1)
-          e <- a3.y3(1)
-        } yield a + b + c + d + e
+  "A StackSafe Tagless algebra should be able to " when {
+    "mapK its Handler to another Handler of the StackSafe" in {
+      import cats.instances.list._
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/test/scala/freestyle/tagless/StacksafeTests.scala
```diff
@@ -263,28 +263,23 @@ class TaglessStacksafeTests extends WordSpec with Matchers {
...
+    "mapK its Handler to another Handler of the StackSafe" in {
+      import cats.instances.list._
+      @tagless @stacksafe trait X {
+        def a: FS[Int]
+      }
+      object Y extends X.Handler[Option] {
+        def a: Option[Int] = Some(42)
+      }
+      val nt: Option ~> List = λ[Option ~> List](_.toList)
+      implicit val hh: X.Handler[List] = Y.mapK(nt)
```
diesalbla, line 276: @loostro Per our Gitter discussion, ¿would this be enough?

ioleo: @diesalbla yeah, this example is exacly what I wanted intially to achieve

peterneyens: @loostro can you share an example where you want `X.Handler` not to be `X` ?

ioleo: @peterneyens I could not get `implicit val logHandlerId: LoggingM.Handler[Id]` to work, after some struggle I found that if I change the type to `implicit val logHandlerId: LoggingM[Id]` it works (see gist), but it feels like a workaround - everywhere in the docs there is Handlers 
https://gist.github.com/loostro/f7b88dfb7dffbabd96a49e37307f262f

peterneyens: If you are just using `@tagless` there is no benefit of using `Handler` as far as I see.

I would always write:
```scala
implicit val summerHandler: Summer[Id] = ...

implicit val logHandlerOption: LoggingM[Option] = ...
```

`Handler` is mostly useful if you want to use an `@tagless` algebra together with `@free` algebras, because it gives you an interpreter for `Foo` (tagless) and `Foo.StackSafe` (free). But I don't think that is a use case that comes up very often.

ioleo: It's fine, as it works without the `Handler`, but then it should be documented somehow.
On the other hand, I guess it does not hurt to override mapK on Handler?

raulraja (approved):
:clap:

PR #530, opened 2018-02-08
Code Refactor

diesalbla:
These are some small changes to the implementation of the macro annotations `@tagless` and `@free`, as well as their `@module`. 

* We extract a shared `Clait` class, which represents the data from either a `Trait` or a `Class` in Scalameta.
* We extract from the four annotations the logic to extract that `Clait` from the annotated source code.
* We move to the `Clait` class the logic to generate or pick the kind-1 parametrer `F[_]`.
* We extract a class of `ErrorMessages`, used almost identically in all macros. 
* We add some logic to the syntax `Ops` classes in `ScalametaUtil`.


juanpedromoreno (approved):
Outstanding job @diesalbla , LGTM

raulraja (approved):
:boom:

PR #533, opened 2018-02-14
FS-532 stacksafe as tagless argument

### LINKED ISSUE #532: Make Stacksafe as an argument of @tagless
A recent issue #517, dealt with in PR #522, introduced a separate `@stacksafe` annotation as a companion to the `@tagless` annotation, and made the Stacksafe pieces of the `@tagless` as optional. 

However, using a separate annotation does not work fine with IntelliJ Idea. Also, it may not make too much sense to have a separate `@stacksafe` annotation that can only be used with `@tagless`. 

To solve this issues, we should make `@stacksafe` into an argument for the `@tagless` macro annotation, so that the code can be rewritten as follows: 

* For a stacksafe and tagless:  `@tagless @stacksafe trait Fox { def fox: FS[Int] }` should become `@tagless(true) trait Fox { def fox: FS[Int] }`
* A non-stacksafe tagless   `@tagless trait Fox { def fox: FS[Int] }` should not need any changes.

diesalbla:
This PR resolves #532. 

We make `stacksafe` into an optional argument of the `@tagless` macro annotation, following the examples [here](http://scalameta.org/paradise/#HowdoIpassanargumenttothemacroannotation?).

### DIFF
--- a/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
+++ b/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
@@ -198,6 +198,18 @@ object ScalametaUtil {
     def addParent( ctorCall: Ctor.Call): Template =
       templ.copy( parents = templ.parents ++ Seq(ctorCall) )
 
+    def addStats( newStats: Seq[Stat]): Template =
+      if (newStats.isEmpty) templ else templ.copy(
+        stats = Some( templ.stats.getOrElse(Seq()) ++ newStats)
+      )
+  }
+
+  implicit class ObjectOps(val obj: Object) extends AnyVal {
+
+    def appendStats(newStats: Seq[Stat]): Object =
+      if (newStats.isEmpty) obj
+      else obj.copy( templ = obj.templ.addStats(newStats))
+
   }
 
 }
--- a/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
+++ b/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
@@ -61,7 +61,6 @@ object ScalametaUtil {
 
     def filtered: Seq[Mod] = mods.filter {
       case mod"@debug" => false
-      case mod"@stacksafe" => false
       case _           => true
     }
 
--- a/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
+++ b/modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
@@ -65,11 +65,6 @@ object ScalametaUtil {
       case _           => true
     }
 
-    def isStackSafe: Boolean = mods.exists {
-      case mod"@stacksafe" => true
-      case _ => false
-    }
-
     def isDebug: Boolean = mods exists {
       case mod"@debug" => true
       case _           => false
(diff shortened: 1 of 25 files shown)


modules/core/shared/src/main/scala/freestyle/tagless/annotations.scala
```diff
@@ -19,13 +19,20 @@ package tagless
...
 
 @compileTimeOnly("enable macro paradise to expand @tagless macro annotations")
-class tagless extends StaticAnnotation {
+class tagless(val stacksafe: Boolean) extends StaticAnnotation {
   import scala.meta._
 
-  inline def apply(defn: Any): Any = meta { taglessImpl.tagless(defn) }
+  inline def apply(defn: Any): Any = meta {
+    val stacksafe: Boolean = this match {
+      case q"new $_(${Lit.Boolean(ss)})" => ss
```
L-Lavigne, line 31: This works for `@tagless(true)` but not for `@tagless(stacksafe = true)`; the compiler accepts both but won't generate the StackSafe tooling in the latter case. Given that naming boolean parameters is generally good practice, we should probably support it.

modules/core/shared/src/main/scala/freestyle/tagless/internal/tagless.scala
```diff
@@ -46,13 +46,11 @@ object taglessImpl {
 
 }
 
-case class Algebra( clait: Clait) {
+case class Algebra( clait: Clait, isStackSafe: Boolean) {
   import clait._
   val errors = new ErrorMessages("@tagless")
   import errors._
 
-  val isStackSafe = clait.mods.isStackSafe
```
L-Lavigne, line 54: The implementation in `ScalametaUtils.ModsOps` still exists; we should probably remove it unless we want to preserve `@stacksafe` support (in that case we need to check for either case in `isStackSafe`).

L-Lavigne:
Good news, this restores IntelliJ macro expansion support. Please see my comments on implementation. Thanks!

modules/core/shared/src/main/scala/freestyle/tagless/annotations.scala
L-Lavigne: Good stuff, the named parameter works in IDEA (though it took more wrangling than usual making the IDE happy - had to actually trigger a syntax error in the tag and then fix it to get it re-parsed ).

modules/core/shared/src/main/scala/freestyle/free/internal/ScalametaUtil.scala
```diff
@@ -65,11 +65,6 @@ object ScalametaUtil {
       case _           => true
     }
 
-    def isStackSafe: Boolean = mods.exists {
-      case mod"@stacksafe" => true
```
L-Lavigne, line 69: Thanks; we still have a reference to `@stacksafe` a few lines above in `filtered()` to remove.

modules/core/shared/src/test/scala/freestyle/tagless/StacksafeTests.scala
```diff
@@ -36,41 +36,55 @@ import modules._
 
 class TaglessStacksafeTests extends WordSpec with Matchers {
 
+  def ok: Unit = (0 shouldEqual 0)
+
   "the @tagless macro annotation, with the extra @stacksafe annotation, should be accepted if it is applied to" when {
```
L-Lavigne: One last reference to `@stacksafe` to remove here

L-Lavigne:
LGTM, thanks!

PR #562, opened 2019-03-13
Update README.md

### LINKED ISSUE #561: Quick start documentation is broken
The documentation is broken at http://frees.io/docs/
Taking the example in the quickstart guide:

```
package meeh

import freestyle.free._
import freestyle.free.implicits._

@free trait Validation {
  def minSize(s: String, n: Int): FS[Boolean]
  def hasNumber(s: String): FS[Boolean]
}

@free trait Interaction {
  def tell(msg: String): FS[Unit]
  def ask(prompt: String): FS[String]
}

@module trait Application {
  val validation: Validation
  val interaction: Interaction

  import cats.implicits._

  def program: FS.Seq[Unit] =
    for {
      userInput <- interaction.ask("Give me something with at least 3 chars and a number on it")
      valid     <- (validation.minSize(userInput, 3), validation.hasNumber(userInput)).mapN(_ && _).freeS
      _         <- if (valid)
        interaction.tell("awesomesauce!")
      else
        interaction.tell(s"$userInput is not valid")
    } yield ()
}

object ApplicationHandler {
  import cats.effect.IO

  implicit val validationHandler = new Validation.Handler[IO] {
    override def minSize(s: String, n: Int): IO[Boolean] = IO{s.size >= n}
    override def hasNumber(s: String): IO[Boolean] = IO{s.exists(c => "0123456789".contains(c))}
  }

  implicit val interactionHandler = new Interaction.Handler[IO] {
    override def tell(s: String): IO[Unit] = IO{println(s)}
    override def ask(s: String): IO[String] = IO { println(s); "This could have been user input 1" }
  }
}

object Main {
  import cats.implicits._
  import cats.effect.IO
  import ApplicationHandler._

  Application.instance.program.interpret[IO].unsafeRunSync
}
```

then sbt/scalac will complain that it cannot find `interpret` for program. It took me quite a while to find out that the type annotation for programm is the reason. If it is rewritten as `def program = for {...` everything works as expected. I do not provide a PR since it would be interesting to know why.

justjoheinz:
Fixes #561

### DIFF
--- a/docs/src/main/tut/docs/README.md
+++ b/docs/src/main/tut/docs/README.md
@@ -18,6 +18,7 @@ To use the project, add the following to your build.sbt:
 
 ```scala
 addCompilerPlugin("org.scalameta" % "paradise" % "3.0.0-M10" cross CrossVersion.full)
+scalacOptions += "-Ypartial-unification"
 ```
 
 [comment]: # (Start Replace)


AntonioMateoGomez (approved):
LGTM

juanpedromoreno (approved):
Thanks for your time @justjoheinz !

